PR #5, opened 2019-01-25
Fix application paths in index.html.

martinpovolny:
Let's see if it fixes the issues with deployment.


martinpovolny:
So the paths in https://github.com/RedHatInsights/topological-inventory-build/blob/ci-stable/index.html are still incorrect.

Something somewhere need to be setup.

PR #12, opened 2019-02-06
Update page classes so we can update PF-next

karelhala:
This will add correct classes for current and future version of PF-next.


karelhala:
ping @martinpovolny

martinpovolny:
Thank you, @karelhala !

PR #15, opened 2019-02-18
Adding of AWS provider type.

martinpovolny:
Add provider type AWS to the "New Source" form.

### DIFF
--- a/src/SmartComponents/ProviderPage/providerForm.js
+++ b/src/SmartComponents/ProviderPage/providerForm.js
@@ -22,7 +22,7 @@ export const providerForm = {
 import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-renderer';
 export const wizardForm = {
     initialValues: {
-        role: 'kubernetes',
+        role: 'kubernetes', // 'aws' for AWS EC2
         verify_ssl: true
     },
     schemaType: 'default',
@@ -37,9 +37,9 @@ export const wizardForm = {
                 name: 'step-1',
                 stepKey: 1,
                 nextStep: {
-                    when: 'source-type',
+                    when: 'source_type',
                     stepMapper: {
-                        aws: 'aws',
+                        amazon: 'amazon',
                         google: 'google',
                         openshift: 'openshift'
                     }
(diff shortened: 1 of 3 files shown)


src/SmartComponents/ProviderPage/providerForm.js
```diff
@@ -22,7 +22,7 @@ export const providerForm = {
 import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-renderer';
 export const wizardForm = {
     initialValues: {
-        role: 'kubernetes',
+        role: 'kubernetes', // 'aws' for AWS EC2
```
agrare: Can this be encoded into the source_type schema ?

src/SmartComponents/ProviderPage/providerForm.js
```diff
@@ -37,9 +37,9 @@ export const wizardForm = {
                 name: 'step-1',
                 stepKey: 1,
                 nextStep: {
-                    when: 'source-type',
+                    when: 'source_type',
```
agrare: What is this case handling that can't be gotten from the source_type records?

agrare:
@martinpovolny the case statements on the hard-coded source_type names are surprising I thought we were trying to get away from that kind of thing?

martinpovolny:
Yes, we where. Therefor I was suggesting to have the creating logic on the server side and on client side.

I will try to first isolate and get rid of any hard-coded provider logic anyway.

src/SmartComponents/ProviderPage/providerForm.js
martinpovolny: This will be gotten from the source_type records in the future.

agrare: Okay can we do this as a follow up now?  I don't think there was any urgency to get this in immediately but could be wrong

src/SmartComponents/ProviderPage/providerForm.js
martinpovolny: I used the initialValues to pass in the role for the Endpoint.

The problem is that the `initialValues` here are for the whole form. Not a single choice of provider. So in order to not have this hard-coded I need to find a different way to get it from the schema.

Until I find such way I added a case with the provider-specific values including the value for the `role`.

PR #17, opened 2019-02-26
Dynamic sources form part 1.

martinpovolny:
Initial steps on making the form sources form generated from the API.

It's problematic if this makes sense because as sources will be registered for several/different apps with different requirements, different fields in the form will be needed.

At that point a more complex logic, one that counts on the different apps will be needed.

### Done

 * load the list of sources
 * generate the combo with list of sources
  * generate pages of the wizzard
  * generate proper sequence of steps/pages

### TODO (another PR)
  * move the form definition to the sources database (entity SourceType) -- first need to add it to https://github.com/martinpovolny/topological_inventory-core

### DIFF
--- a/src/SmartComponents/ProviderPage/providerForm.js
+++ b/src/SmartComponents/ProviderPage/providerForm.js
@@ -1,25 +1,5 @@
-export const providerForm = {
-    schemaType: 'mozilla',
-    schema: {
-        title: 'Add an Openshift Provider',
-        type: 'object',
-        properties: {
-            name: { title: 'Provider Name', type: 'string' },
-            description: { title: 'Description', type: 'string' },
-            url: { title: 'URL', type: 'string' },
-            verify_ssl: { title: 'Verify SSL', type: 'boolean', default: false },
-            user: { title: 'User Name', type: 'string', default: '' },
-            token: { title: 'Token', type: 'string', default: '' },
-            password: { title: 'Password', type: 'string', minlength: 6 }
-        },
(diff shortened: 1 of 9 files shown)


martinpovolny:
To fix the (now hard-coded) calculation of entry point name for the created source, I would need this functionality: https://github.com/final-form/react-final-form/issues/387

src/SmartComponents/ProviderPage/providerForm.js
```diff
@@ -1,141 +1,138 @@
...
+            type: validatorTypes.REQUIRED
+        }]
+    }]
+});
+
+const summaryStep = () => ({
+    fields: [{
+        name: 'summary',
+        component: 'summary',
+        assignFieldProvider: true
```
Hyperkid123: `assignFieldProvider` is no longer necessary. FieldProvider is passed to each component by default.

src/SmartComponents/ProviderPage/providerForm.js
```diff
@@ -1,141 +1,138 @@
...
+                )
             }]
-        }]
-    },
-    uiSchema: {
-    //password: {'ui:widget': 'password'}
-    }
-};
+        },
+        uiSchema: {
```
Hyperkid123: You can get rid of this key entirely

martinpovolny:
@Hyperkid123 : I addressed your feedback and added the possibility to use an environment variable for setting the BASE_PATH.

I'd like to do next steps in a new PR.

PR #22, opened 2019-03-11
Use plain fetch to get source_types.

martinpovolny:
There's something broken with the javascript API client.

Using plain `fetch` for now.


Hyperkid123:
I would probably leave the API clients for now until we have some stable version. I am having troubles with it myself in catalog UI. I am constantly switching between the client and direct calls.

PR #38, opened 2019-04-03
Source Editing

martinpovolny:
### Done
  * route added
  * separated Modal... as a component
  * loading data 
  * form schema for editing
  * put data in the initialValues
  * saving
  * updated dependencies

### Todo (can be a separate PR)
   * check username
   * get a fix for https://github.com/patternfly/patternfly-react/issues/1711


Hyperkid123:
There is a missing dependency for `'redux-mock-store` for tests.

PR #50, opened 2019-04-12
Temporary limit sorting to "Name".

martinpovolny:
ping @Hyperkid123


src/components/SourcesFilter.js
```diff
@@ -11,7 +11,7 @@ class SourcesFilter extends Component {
         <SimpleTableFilter
             buttonTitle={null}
             placeholder='Filter by source name'
-            options={{
+            xoptions={{
```
Hyperkid123: Is this supposed to be `xoptions`?

martinpovolny: Yes, I am disabling it (temporary) by renaming the option.

PR #52, opened 2019-04-16
Split and style OpenShift wizard page.

martinpovolny:
### OpenShift page work

 * add support for multiple pages per `SourceType`
 * OpenShift page split into two and
 * restyled with
 * a new component for the SSL label with popover help.

osst1

osst2

osst3

### DIFF
--- a/src/SmartComponents/ProviderPage/providerForm.js
+++ b/src/SmartComponents/ProviderPage/providerForm.js
@@ -323,21 +323,16 @@ const initialValues = source => {
     };
 };
 
-const logIt = (msg, foo) => {
-    console.log(msg, foo);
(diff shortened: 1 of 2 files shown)


src/SmartComponents/ProviderPage/providerForm.js
```diff
@@ -290,6 +323,11 @@ const initialValues = source => {
     };
 };
 
+const logIt = (msg, foo) => {
+    console.log(msg, foo);
```
Hyperkid123: Is this necessary?

martinpovolny: no --> removed

PR #53, opened 2019-04-17
Add summary to the add/edit wizard.

martinpovolny:
summary

### TODO

 * ~~connect state, display proper provider label~~ (done)

### DIFF
--- a/src/SmartComponents/ProviderPage/providerForm.js
+++ b/src/SmartComponents/ProviderPage/providerForm.js
@@ -282,7 +282,11 @@ const summaryStep = () => ({
         component: 'summary'
     }],
     stepKey: 'summary',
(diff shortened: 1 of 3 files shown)


nlcwong:
Do not use a disabled input text box as the component, you should be using TextContent.

martinpovolny:
> Do not use a disabled input text box as the component, you should be using TextContent.

As I wrote you in a private message as a response to this very sentense:

Once I know what exact variants of TextContent and how I should use, I'll do that. I'll replace the markup used in this PR.

Can you tell me what are the TextContent elements I should use for

 1. the labels,
 2. the fields?

If you know, please tell me.

Untill I have such information I am moving forward with disabled form fields.

PR #56, opened 2019-04-22
Switch to using sources-api service

agrare:
Point the sources-ui to the sources-api service instead of the topological-inventory-api


agrare:
This will need a sources-api js client

martinpovolny:
Seems the changes your did are the same I did and actually I was later to the show. But I noticed your PR just now. Sorry and Thx!

PR #57, opened 2019-04-23
Sources api v1

martinpovolny:
### TODO:
 * ~~source creating is now returning `{message: "Unauthorized"}`~~ (fixed in https://github.com/ManageIQ/sources-api/pull/17)


agrare:
@Hyperkid123 I think this should not have been merged, since it was a workaround for something which was already fixed.

Hyperkid123:
Do you mean the tenants? This one removes it.

agrare:
Yeah sorry read the diff too quickly

PR #58, opened 2019-04-23
Use new API client

martinpovolny:
Move to the new API client by @karelhala and @Hyperkid123 

to be rebased after https://github.com/ManageIQ/sources-ui/pull/57 is merged


src/api/entities.js
```diff
@@ -1,25 +1,24 @@
...
-export function getApiInstance() {
+export function getSourcesApi() {
     if (apiInstance) {
         return apiInstance;
     }
 
-    apiInstance = new TopologicalInventory.DefaultApi();
-    let defaultClient = TopologicalInventory.ApiClient.instance;
-    defaultClient.basePath = TOPOLOGICAL_INVENTORY_API_BASE;
+    apiInstance = new SourcesDefaultApi(undefined, SOURCES_API_BASE, axiosInstance);
```
karelhala: You might want to use response interceptor to return data instead of object with data
```
instance.interceptors.response.use((response) => {
  return response.data;
});
```

martinpovolny:
@karelhala, @Hyperkid123 : I updated the PR based on your feedback. Please, check.

Thx!

PR #60, opened 2019-04-24
UX quick wins

martinpovolny:
Various visual fixes according to UX review.


src/views/sourcesViewDefinition.js
```diff
@@ -13,7 +13,7 @@ export const sourcesViewDefinition = {
         { title: 'Name', value: 'name', searchable: true },
         { title: 'Type', value: 'source_type_id', searchable: false, formatter: 'sourceTypeFormatter' },
         // this column does not actually exist in the API, but is required by the design;
-        { title: 'Application', value: 'applications', searchable: false },
+        { title: null, value: 'applications', searchable: false },
```
Hyperkid123: I would probably just delete the `title` key. JS has some bad habits when it comes to `null` vs `undefined`.

martinpovolny: well I have it consistent with the lines above (see id, uid, tenant_id)

so far no problems

PR #71, opened 2019-05-22
Make endpoint `default` when creating a source.

martinpovolny:


agrare:
@martinpovolny can this be overridden e.g. when adding more than one endpoint we only want one to be default?

martinpovolny:
There's no way to add more that one endpoint in the UI.

I have no such design (wireframes) nor requirements.

Should there be one?

martinpovolny:
Also I do think that this shold be handled by the backend. That is: if there's just one endpoint, make it default by default. That is a more robust behavior that the current one.

Do you want this PR merged?

agrare:
Eventually we will need to be able to add more than one endpoint for some providers (e.g. Prometheus endpoint for openshift metrics or db connection info for rhev metrics) and if there is more than one then the backend can have no way of knowing which is the "main" one.  It is probably something we'll have to encode in the json schema definition for the provider.

PR #75, opened 2019-05-23
Translate all strings into react-intl components

rvsia:
+ Plus small refactoring

+ After a discussion with @Hyperkid123 moved translation of delete notification into the component and sending the translated message as a second argument

TODO
- [x] Schemas


rvsia:
@martinpovolny

rvsia:
@martinpovolny Everything (except the pagination, which is missing options how to do it) should be in translatable form. :earth_americas:  Can you please test in UI if everything is working? Thanks.

PR #81, opened 2019-07-25
Use new wizard from frontend components

rvsia:
**Description**

- use https://github.com/RedHatInsights/frontend-components/tree/master/packages/sources as a Add source wizard
=> updates PF4 dependencies
=> updates Edit wizard (full remake in the future)
=> updates Checkbox API

- removes all functionality around createSource
=> removes unused components which use it

**Before**

Screenshot from 2019-07-25 15-09-34

Screenshot from 2019-07-25 15-09-42

**After**

Screenshot from 2019-07-25 14-58-10

Screenshot from 2019-07-25 14-58-05


rvsia:
@martinpovolny @Hyperkid123

Please, review! :+1: 

What is not in this PR:
- UI refinements, modified schema, etc.
- Better wording, etc.
- Any tests

This PR is just for using a external Add Source wizard (so, no double code) instead of the old one, removing unnecessary files and updating Edit Source to newest PF with minimum changes (because we can't use new and old wizard together). (Editing does not work now, please see JIRA: [TPINVTRY-534](https://projects.engineering.redhat.com/browse/TPINVTRY-534) )

cc @terezanovotna

src/api/entities.js
```diff
@@ -48,21 +47,14 @@ export const doLoadEndpoints = sourceList =>
     axiosInstance.get(`${SOURCES_API_BASE}/endpoints/?source_id=${sourceList}`);
 
 export function doRemoveSource(sourceId) {
-    return getSourcesApi().deleteSource(sourceId).then((sourceDataOut) => {
-        console.log('API call deleteSource returned data: ', sourceDataOut);
-    }, (_error) => {
-        console.error('Source removal failed.');
+    return getSourcesApi().deleteSource(sourceId).catch((_error) => {
```
Hyperkid123: I think we should not be hiding the error. The notification component can handle that.

src/components/Description.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Description = ({ name, content, className }) => <div className={ className } id={ name }>
```
Hyperkid123: add
(
 <div...
)

And we should add a linter rule for this.

rvsia: The rule was added! :straight_ruler:

src/api/entities.js
rvsia: Created :+1:

PR #83, opened 2019-08-07
Updates Add Source wizard

rvsia:
**Description**

- updates Add Source wizard to allow selecting of an application
- updates API call to get Sources (instead of older and now not-working API format graphQl is used) => application are correctly displayed in the table
- removes unnecessary lodash functions across used pages
- faster loading of the table because all entities are being loaded at same time

see https://github.com/RedHatInsights/frontend-components/pull/114

@martinpovolny @Hyperkid123 

image

### DIFF
--- a/src/components/SourcesSimpleView.js
+++ b/src/components/SourcesSimpleView.js
@@ -7,13 +7,11 @@ import { Table, TableHeader, TableBody, sortable } from '@patternfly/react-table
 import { TextContent, Text, TextVariants } from '@patternfly/react-core';
 import { injectIntl } from 'react-intl';
 
-import flatten from 'lodash/flatten';
-import filter from 'lodash/filter';
 import ContentLoader from 'react-content-loader';
 import moment from 'moment';
 
 import SourceExpandedView from './SourceExpandedView';
-import { loadEntities, selectEntity, expandEntity, removeSource, sortEntities } from '../redux/actions/providers';
+import { loadEntities, expandEntity, sortEntities } from '../redux/actions/providers';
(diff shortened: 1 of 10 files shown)


src/components/SourcesSimpleView.js
```diff
@@ -87,11 +83,13 @@ class SourcesSimpleView extends React.Component {
         ]
     );
 
-    sourceIsOpenShift = source => this.sourceTypeMap.get(source.source_type_id) === 'openshift';
+    sourceIsOpenShift = source => this.props.sourceTypes.find((type) => type.id === source.source_type_id).name === 'openshift';
```
Hyperkid123: This will crash in runtime

src/components/SourcesSimpleView.js
```diff
@@ -87,11 +83,13 @@ class SourcesSimpleView extends React.Component {
...
     formatURL = source => source.endpoints && source.endpoints[0] && endpointToUrl(source.endpoints[0]);
 
-    applicationFormatter = apps => apps.map(a => this.appTypeMap.get(a.application_type_id)).join(', ');
-    sourceTypeFormatter = sourceType => (this.sourceTypeMap.get(sourceType) || sourceType || '');
+    applicationFormatter = apps =>
+        (apps.map(a => this.props.appTypes.find((type) => type.id === a.application_type_id).display_name).join(',\n'));
```
Hyperkid123: `this.props.appTypes.find((type) => type.id === a.application_type_id).display_name` will crash if nothing is found.

PR #89, opened 2019-08-14
Fix missing attribute to display source type

rvsia:
Something I missed the all time

**Before**

image

**After**

image

@miq-bot add_label


rvsia:
@miq-bot add_label bug

PR #90, opened 2019-08-21
Update PF react to latest

rvsia:
**Description**

Just updates PF4 to latest, it will solve issue with popover

image

@Hyperkid123


Hyperkid123:
@rvsia while you are at it i think you should add some spacing between the icon and the text.

rvsia:
image

PR #95, opened 2019-09-06
Add loader for deleting source

rvsia:
**Description**

- Adds loader for deleting a source.
- Also the source is deleted from the application state, so no more refreshing the whole list.

**Before**

deletebefore

**Success**

successdelete

**Error**

error

TODO
- [x] remove double errors
- [x] refactor the list
- [x] Tests

@terezanovotna

### DIFF
--- a/src/api/entities.js
+++ b/src/api/entities.js
@@ -36,8 +36,8 @@ export const doLoadAppTypes = () =>
     axiosInstance.get(`${SOURCES_API_BASE}/application_types`);
 
 export function doRemoveSource(sourceId) {
-    return getSourcesApi().deleteSource(sourceId).catch((_error) => {
-        throw { error: 'Source removal failed.' };
+    return getSourcesApi().deleteSource(sourceId).catch((data) => {
+        throw { error: { detail: data.data.errors[0].detail } };
     });
 }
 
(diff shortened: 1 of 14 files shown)


terezanovotna:
I like it!

rvsia:
@Hyperkid123

src/api/entities.js
```diff
@@ -112,14 +112,14 @@ export function doUpdateSource(source, formData) {
             .then((authenticationDataOut) => {
                 return authenticationDataOut;
             }, (_error) => {
-                throw { error: 'Authentication update failure.' };
+                throw { error: { title: 'Authentication update failure.' } };
```
Hyperkid123: Its not very good idea to completly discard the error response. If you dont want to show it as notification a least display it in console.

src/redux/actions/providers.js
```diff
@@ -82,17 +82,20 @@ export const updateSource = (source, formData, title, description) => (dispatch)
...
 
-export const removeSource = (sourceId, title) => (dispatch) =>
-    doRemoveSource(sourceId).then(_finished => dispatch({
-        type: ADD_NOTIFICATION,
-        payload: {
-            variant: 'success',
-            title
+export const removeSource = (sourceId, title) => ({
+    type: ACTION_TYPES.REMOVE_SOURCE,
+    payload: () => doRemoveSource(sourceId),
```
Hyperkid123: The promise middleware was deliberately removed in earlier versions of this repo. Are you sure you want it Back?

src/api/entities.js
rvsia: This was just to preserve compatibility with new notification settings, the whole edit form should be changed at some time in the future :skull_and_crossbones: but I changed it

src/redux/actions/providers.js
rvsia: Was there any specific reason? I think it serves its purpose, the code is clean and is officially supported by notifications middleware.

Hyperkid123: It was getting in a way for more complex requests. That is it. You can still use the notification middleware without promise middleware.

PR #102, opened 2019-09-23
Limit apps selection when managing apps

rvsia:
**Description**

Only supported apps are available when managing apps for a source.

+ removes loading when removing an app from a source
+ when creating an app, the app is stored directly into redux, so no need to refresh all data

**Before**

beforeappremove

**After**

Success

removeAppSucc

Error

appremoveerror

TODO
- [x] tests

@terezanovotna


rvsia:
@martinpovolny

rvsia:
@miq-bot remove_label unmergeable

terezanovotna:
Looks good!

PR #105, opened 2019-09-30
Do not show empty/default ports

rvsia:
https://projects.engineering.redhat.com/browse/TPINVTRY-697

**Description**

- Hide empty/default ports in the table/edit form
- little refactoring (using `Promise.all` for faster loading times)
- name validator in edit form (used from `frontend-components-sources`)
- hide `Application` in the edit summary

**Before**

List

image

Undefined/Default

image

**After**

List

image

Undefined/Default

image


rvsia:
@Hyperkid123  please review

(The edit form has still some UX issues (loading after submit, etc.) and it should be completely changed.. so do not focus on these :smile_cat: )

PR #109, opened 2019-10-01
Add graphql axios interceptor

rvsia:
**Description**

Adds a graphql interceptor to Axios to return an error message

**Before**

image

**After**

image


rvsia:
@Hyperkid123

PR #132, opened 2019-11-11
Add undo button for cancelling add source

rvsia:
https://projects.engineering.redhat.com/browse/TPINVTRY-734

**Description**

add undo button if user cancel adding by a mistake

undoButton

- when values are empty, there is no notification
- notifications stacks up (so user can return multiple time)
- the notification is dismissed after the undo

TODO
- [x] tests
- [x] undo when on add source wizard


rvsia:
@Hyperkid123

rvsia:
@terezanovotna

src/components/UndoButton/UndoButtonAdd.js
```diff
@@ -0,0 +1,61 @@
...
+    history,
+    removeMessage,
+    values,
+    undoAddSource
+}) => (
+    <Button variant="link" isInline onClick={() => {
+        const notification = notifications.find(({ customId }) => customId === messageId);
+
+        if (notification) {
+            removeMessage(notification.id);
```
Hyperkid123: So the notifications middleware allows clearing all notifications and there is specific action for it. Maybe you could do that instead of finding the notification? Don't know if closing all is useful here

rvsia: User can have a stack of notifications.

(And it allows to stack multiple undoes :D )

terezanovotna:
Very nice!

PR #137, opened 2019-11-13
Use get base name

rvsia:
**Description**
- used `getBaseName` from the shared components

And also
- ~~deleted `entry-dev` as it is same as the `entry`, is that right?~~

@karelhala @Hyperkid123


rvsia:
@karelhala @Hyperkid123 updated, no deletion :sob:

PR #139, opened 2019-11-15
Update spandx to latest changes in navigation

rvsia:
**Description**

No hybrid anymore. :( 

@karelhala @Hyperkid123

### DIFF
--- a/config/spandx.config.js
+++ b/config/spandx.config.js
@@ -6,8 +6,7 @@ const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docke
 
 module.exports = {
     routes: {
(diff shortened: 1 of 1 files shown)


config/spandx.config.js
```diff
@@ -7,7 +7,7 @@ const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docke
 module.exports = {
     routes: {
         '/insights/settings/sources': { host: `http://${localhost}:8002` },
```
karelhala: This one is no longer valid, but you can keep i here. If you want to.

rvsia: :+1:

PR #165, opened 2019-11-29
Add availability status column

rvsia:
https://projects.engineering.redhat.com/browse/TPINVTRY-735

@terezanovotna

image


Hyperkid123:
@rvsia the pop over is nit aligned. Can you fix that so the root for the popover is the text not the table cell?

rvsia:
@Hyperkid123 updated! See screenshot in description

PR #168, opened 2019-12-02
Support adding application to existing source

rvsia:
Allow to attach apps to existing sources.

Imported sources
- only selection of the app

Sources with credentials
- credentials are pre-filled when app uses them (except the password, which is not required, same as edit)

Sources without credentials
- same flow as adding

Todo:
- [x] https://github.com/RedHatInsights/frontend-components/pull/291
  - [x] tests
- [x] submitting
- [x] tests
  - [x] authtypesetter
  - [x] schema generator
  - [x] add application
- [x] memo application selection when source is changed
- [x] make password no requried (because we cannot get these values by API and prefilled it)
- [x] missing values when adding more than one app

attachingapp


src/api/doAttachApp.js
```diff
@@ -0,0 +1,137 @@
+import { addedDiff, updatedDiff } from 'deep-object-diff';
+import { patchSource, handleError } from '@redhat-cloud-services/frontend-components-sources';
+import isEmpty from 'lodash/isEmpty';
+import merge from 'lodash/merge';
+import cloneDeep from 'lodash/cloneDeep';
+
+import { getSourcesApi, doCreateApplication } from './entities';
+import { urlOrHost } from './doUpdateSource';
+
+export const convertToUndefined = (value) => {
```
Hyperkid123: This might not be necessary after : https://github.com/data-driven-forms/react-forms/pull/273
I think it can be merged soon

rvsia: Yeah, I am planning to use it in the future, but right now I would rather move this forward. :incoming_envelope:

src/components/AddApplication/AuthTypeSetter.js
```diff
@@ -0,0 +1,63 @@
...
+
+export const innerSetter = ({ sourceTypes, source, appTypes, formOptions, checkAuthType, authenticationValues }) => {
+    const sourceType = sourceTypes.find(({ id }) => id === source.source_type_id);
+
+    const formValues = formOptions.getState();
+
+    const application = appTypes.find(
+        ({ id }) => id === get(formValues, 'values.application.application_type_id', undefined)
+    );
+    const supported_auth_type = application ? application.supported_authentication_types[sourceType.name][0] : '';
```
Hyperkid123: what if `application.supported_authentication_types[sourceType.name][0] ` does not exist?

rvsia: Fixed

PR #177, opened 2019-12-10
Send to endpoint API when editing url

rvsia:
@slemrmartin 

This should fix missing Endpoint's API call when editing URL values.


Hyperkid123:
@rvsia eslint :wink:

PR #192, opened 2019-12-13
Sort SourceTypes in filter alphabetically

rvsia:
**Before**

image

**After**

image

@terezanovotna


terezanovotna:
Very nice!

PR #206, opened 2020-01-08
Use useParams in useSource hook

rvsia:
`useSource` does not need ID props anymore (but it is possible to use it). It takes the `id` right away from url via the `useParams` hook.

### DIFF
--- a/src/hooks/useSource.js
+++ b/src/hooks/useSource.js
@@ -1,7 +1,9 @@
 import { useSelector } from 'react-redux';
+import { useParams } from 'react-router-dom';
 
-export const useSource = (id) => {
-    const source = useSelector(({ providers }) => providers.entities.find(source => source.id  === id));
+export const useSource = () => {
+    const { id } = useParams();
+    const source = useSelector(({ providers }) => providers.entities.find(source => source.id === id));
 
     return source;
 };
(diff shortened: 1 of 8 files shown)


src/hooks/useSource.js
```diff
@@ -1,7 +1,9 @@
 import { useSelector } from 'react-redux';
+import { useParams } from 'react-router-dom';
 
-export const useSource = (id) => {
-    const source = useSelector(({ providers }) => providers.entities.find(source => source.id  === id));
+export const useSource = (sourceId) => {
```
Hyperkid123: Firstly. Nice work on the hook. Have one comment here. I think that it would be safer and less magical to make the `sourceId` required param and ditch the `useParam`. You are playing with potential side effects which might bite you later on.

rvsia: That's the point of this PR. I was using `useParams` and `useSource` in all components together, so this PR just removes unnecessary duplication across the components (and less imports). 

Right now, the purpose of `useSource` is to get the active source (which is being edited, removed, etc.) by url... actually, the `sourceId` is not used anywhere. :D

Hyperkid123: Ok so in that case, I would remove the param and have only one source of the ID.

rvsia: Okay, will do it!

rvsia: Removed

PR #207, opened 2020-01-08
Updates PFcore and others

rvsia:
- Manual update of PF Core package (+ all others as well)
- One test had to be updated because of changes in PF Core


src/test/components/SourcesSimpleView/formatters.spec.js
```diff
@@ -410,7 +411,7 @@ describe('formatters', () => {
                 const wrapper = mount(wrapperWithIntl(availabilityFormatter('some nonsense', SOURCE, { appTypes: APPTYPES })));
 
                 expect(wrapper.text().includes('--')).toEqual(true);
-                expect(wrapper.find(Popover).text().includes('No application')).toEqual(true);
+                expect(ReactDOMServer.renderToStaticMarkup(wrapper.find(Popover).props().bodyContent)).toEqual('<h1>No application connected.</h1>');
```
Hyperkid123: WOW where did this come from?

rvsia: Popover is now using portal to show things and I need to check what is inside when you open it.. so the most convenient way I found was just to render the content of the popover to static markup and compare it. :D simulate('click') didn't work for me.

Hyperkid123: Ah. Portal it is. Now it makes sense.

PR #229, opened 2020-01-22
Update all deps

rvsia:


Hyperkid123:
@rvsia linter is dead :skull:

PR #234, opened 2020-01-28
Refactor routes to use one source of truth for all functionality

rvsia:
Introduces one source of truth for all components' shared functionality
```jsx
export const routes = {
    sources: {
        path: '/'
    },
    sourcesNew: {
        path: '/new',
        writeAccess: true
    },
    sourcesEdit: {
        path: '/edit/:id',
        writeAccess: true
    },
    sourcesRemove: {
        path: '/remove/:id',
        redirectNoId: true,
        writeAccess: true
    },
    sourceManageApps: {
        path: '/manage_apps/:id',
        redirectNoId: true,
        writeAccess: true
    }
};
```

Then, CustomRoute component handle all these params.

```jsx
<CustomRoute exact route={routes.sourcesRemove} Component={SourceRemoveModal}/>
```

### DIFF
--- a/src/components/CustomRoute/CustomRoute.js
+++ b/src/components/CustomRoute/CustomRoute.js
@@ -7,7 +7,7 @@ import { useSource } from '../../hooks/useSource';
 import RedirectNoId from '../RedirectNoId/RedirectNoId';
 
 const CustomRouteInternal = ({ route, children }) => {
-    const source = useSource();
+    const source = route.redirectNoId ? useSource() : undefined;
 
     if (!source && route.redirectNoId) {
         return <RedirectNoId />;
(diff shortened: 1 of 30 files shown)


src/components/CustomRoute/CustomRoute.js
```diff
@@ -0,0 +1,57 @@
...
+        path: PropTypes.string.isRequired,
+        redirectNoId: PropTypes.bool,
+        writeAccess: PropTypes.bool
+    }).isRequired,
+    children: PropTypes.node.isRequired
+};
+
+const CustomRoute = ({ route, componentProps, Component, ...props }) => {
+    return (
+        <Route
```
Hyperkid123: extra block. Also, you can use children instead of `render` prop.

PR #235, opened 2020-01-29
Trigger availability status check after a creation of source

rvsia:
Fixes TPINVTRY-740


rvsia:
@Hyperkid123 I suppose that the only UI knows when all endpoints are successfully called. You can have a source which is selfcontained, a source with an endpoint or you can wait until the record is created in the Cost Management app... so the backend does not really know when it should trigger the check. Am I right, @gtanzillo

PR #238, opened 2020-01-31
Fix trailing slash in the root url

rvsia:
`/beta/settings/sources/` > `/beta/settings/sources`

Also, paths are always used as constants, so it makes really simple to change it on one place.


src/test/app.spec.js
```diff
@@ -87,7 +87,7 @@ describe('App spec js', () => {
         expect(wrapper.find(Main)).toHaveLength(1);
         expect(wrapper.find(Routes)).toHaveLength(1);
         expect(wrapper.find(Router)).toHaveLength(1);
-        expect(wrapper.find(Router).props().basename).toEqual('//');
```
Hyperkid123: 😄

PR #241, opened 2020-01-31
Make top pagination compact

rvsia:
Be consistent with other apps

**Before** 

image

Loading

image

**After**

image

Loading

image


Hyperkid123:
@rvsia do is there bottom pagination?

PR #264, opened 2020-02-13
Copy code from utils to achieve smaller size

rvsia:
Exports from frontend components are too large, so I just copied interceptors and the one helper method I am using. App is now 130 KB smaller.

@Hyperkid123 @karelhala 

**Before**

image
image

**After**

image
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -3,13 +3,13 @@ import { NotificationsPortal } from '@redhat-cloud-services/frontend-components-
 import { Main } from '@redhat-cloud-services/frontend-components/components/Main';
 import { IntlProvider } from 'react-intl';
 import { BrowserRouter as Router } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -3,13 +3,13 @@ import { NotificationsPortal } from '@redhat-cloud-services/frontend-components-
 import { Main } from '@redhat-cloud-services/frontend-components/components/Main';
 import { IntlProvider } from 'react-intl';
 import { BrowserRouter as Router } from 'react-router-dom';
-import { getBaseName } from '@redhat-cloud-services/frontend-components-utilities/files/helpers';
```
karelhala: It might be better to split helpers file into multiple smaller ones.

rvsia: If it would be smaller than 595 bytes... :smile:

Hyperkid123: > It might be better to split helpers file into multiple smaller ones.

I concur

PR #265, opened 2020-02-18
Update almost all deps

rvsia:
* eslint needed better configuration
* removing mini-css extract plugin
* adds object as proprype to custom component because of lazy importing
* switched to devServer (bye hotloading :( )
* in action resolver buttons are used as components


rvsia:
@Hyperkid123 this should be ready to review

src/components/CustomRoute/CustomRoute.js
```diff
@@ -45,7 +45,7 @@ CustomRoute.propTypes = {
         writeAccess: PropTypes.bool
     }).isRequired,
     componentProps: PropTypes.any,
-    Component: PropTypes.func.isRequired
+    Component: PropTypes.oneOfType([PropTypes.func, PropTypes.object]).isRequired
```
Hyperkid123: I dont think this should be `object`. Did you try `element`?

rvsia: I think it should be an object because it gets an object, when the `Component` is not loaded yet. PropWarning starts to appear after I used lazy imports. Element throws an error.

PR #277, opened 2020-03-03
Hide topology invetory from available apps

rvsia:
TPINVTRY-878

- [x] Wizard https://github.com/RedHatInsights/frontend-components/pull/385
- [x] Table - should stay

- [x] Remove app ?
- [x] Remove source ? 
^^  these two should be no problem in GA, as it won't be possible to attach TOP INV to a source


slemrmartin:
@rvsia so it'll completely dissapear? Also fro CI/QA? So I won't be no longer able to test anything regarding topological inventory through UI?

rvsia:
@slemrmartin Yeah, in UI you won't be able to add TOP INV or attach it to a source. However, you can still see it in the list or remove it.

rvsia:
@slemrmartin However, as TOP INV does not nedd any special fields, you can create a source with no app and with needed authentications in the UI and then made a simple API call to create the application.

PR #282, opened 2020-03-06
Always show scrollbar

rvsia:
Shrinking of the table is so disrupting.

Do you think there is some better solution, @Hyperkid123 ?

**Before**

before

**After**

after

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -100,3 +100,7 @@
   width: 32px;
   height: 32px;
 }
+
+.pf-l-page__main {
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@rvsia I personally don't think that is an issue. It's mostly browser/OS dependent. On OSX you have hiding scrollbar, the same goes for edge (don't know about Firefox). This is just a natural part of browsers.

rvsia:
@Hyperkid123 I hate it, but okay :D

PR #285, opened 2020-03-12
Create authApp entity when attaching an app

rvsia:
- [x] tests

### DIFF
--- a/src/api/doAttachApp.js
+++ b/src/api/doAttachApp.js
@@ -60,6 +60,8 @@ export const doAttachApp = async (values, formApi, authenticationInitialValues, 
 
         if (filteredValues.source && !isEmpty(filteredValues.source)) {
             promises.push(getSourcesApi().updateSource(sourceId, filteredValues.source));
+        } else {
+            promises.push(Promise.resolve(undefined));
         }
 
         const hasModifiedEndpoint = filteredValues.endpoint && !isEmpty(filteredValues.endpoint);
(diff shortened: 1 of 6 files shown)


src/api/doAttachApp.js
```diff
@@ -60,6 +60,8 @@ export const doAttachApp = async (values, formApi, authenticationInitialValues,
 
         if (filteredValues.source && !isEmpty(filteredValues.source)) {
             promises.push(getSourcesApi().updateSource(sourceId, filteredValues.source));
+        } else {
+            promises.push(Promise.resolve(undefined));
```
Hyperkid123: what is the purpose of these? Do you just need the correct order of arguments in promise.all?

src/api/entities.js
```diff
@@ -47,7 +47,8 @@ export const getSourcesApi = () => ({
     postGraphQL: (data) => axiosInstanceInsights.post(`${SOURCES_API_BASE}/graphql`, data),
     listSourceTypes: () => axiosInstanceInsights.get(`${SOURCES_API_BASE}/source_types`),
     doLoadAppTypes: () => axiosInstanceInsights.get(`${SOURCES_API_BASE}/application_types`),
-    deleteApplication: (id) => axiosInstanceInsights.delete(`${SOURCES_API_BASE}/applications/${id}`)
+    deleteApplication: (id) => axiosInstanceInsights.delete(`${SOURCES_API_BASE}/applications/${id}`),
+    createAuthApp: (data) => axiosInstanceInsights.post(`${SOURCES_API_BASE_V2}/application_authentications`, data)
```
Hyperkid123: I probably know the answer but v1 and v2 are not compatible?

src/api/doAttachApp.js
rvsia: I need to extract info from responses, so I have to know indexes of all promises. And because they are dynamically added, I have to fix their positions by using `undefined`

src/api/entities.js
rvsia: :smirk:

PR #295, opened 2020-03-24
Fix warnings in tests

rvsia:
All remaining errors/warning in the test are caused by data driven forms and will be fixed in version 2


Hyperkid123:
Travis is for some reason not updating GH even though the job is finished. Let us wait a bit.

PR #301, opened 2020-03-31
Allow to select the autentication when adding app

rvsia:
- Allows to use already existing authentication (multiple authentication of the same type) when adding an app 
- when there is no available authentication, nothing changed

chooseauthentication

TODO:
- [x] change dontRequirePassword to better solution https://github.com/RedHatInsights/frontend-components/pull/428
- [x] remove authentication when changed app type
- [x] show which applications are using the authentications
- [x] tests

### DIFF
--- a/src/components/AddApplication/AddApplicationSchema.js
+++ b/src/components/AddApplication/AddApplicationSchema.js
@@ -85,8 +85,12 @@ const selectAuthenticationStep = ({ source, authenticationValues, sourceType, ap
                 const appTypeId = app ? app.application_type_id : '';
                 const appType = appTypeId ? applicationTypes.find(({ id }) => id === appTypeId) : '';
 
+                const includeUsername = values.username ? `-${values.username}` : '';
+                const includeAppName = appType ? `-${appType.display_name}` : `-unused-${values.id}`;
+                const label = `${supportedAuthTypeName}${includeUsername}${includeAppName}`;
+
                 return {
-                    label: `${supportedAuthTypeName}-${appType ? `-${appType.display_name}` : `unused-${values.id}`}`,
+                    label,
                     value: values.id,
                 };
             });
--- a/src/components/AddApplication/AddApplicationSchema.js
+++ b/src/components/AddApplication/AddApplicationSchema.js
@@ -54,7 +54,9 @@ SelectAuthenticationDescription.propTypes = {
 export const hasAlreadySupportedAuthType = (authValues = [], appType, sourceTypeName) =>
     authValues.find(({ authtype }) => authtype === get(appType, `supported_authentication_types.${sourceTypeName}[0]`));
 
-const selectAuthenticationStep = ({ authenticationValues, sourceType, applicationTypes, modifiedValues }) => {
+const selectAuthenticationStep = ({ source, authenticationValues, sourceType, applicationTypes, modifiedValues }) => {
+    console.log(source);
+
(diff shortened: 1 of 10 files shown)


src/components/AddApplication/AddApplicationSchema.js
```diff
@@ -31,32 +33,140 @@ export const ApplicationSummary = () => (<TextContent>
...
+        const supportedAuthType = get(app, `supported_authentication_types[${sourceType.name}][0]`, '');
+
+        if (supportedAuthType && hasAlreadySupportedAuthType(authenticationValues, app, sourceType.name)) {
+            const supportedAuthTypeName =
+                get(sourceType, `schema.authentication`, {}).find(({ type }) => type === supportedAuthType).name;
+
+            const radioOptions = authenticationValues.filter(({ authtype }) => authtype === supportedAuthType).map((values) => {
+                const app = source.applications.find(({ authentications }) => authentications.find(({ id }) => id === values.id));
+
+                const appTypeId = app ? app.application_type_id : '';
```
karelhala, line 86: No need to assign this to a value
```JS
const appType = applicationTypes.find(({ id }) => id === app && app.application_type_id) ||
  {
    display_name: `unused-${values.id}`
  }
const includeUsername = values.username ? `-${values.username}` : '';

  return {
    label: `${supportedAuthTypeName}${includeUsername}-${appType.display_name}`
  }
```

You could simplify it even more by directly calculating `includeUsername` and `appType` when creating label but that is kinda hard to read.
```JS
return {
    label: `${supportedAuthTypeName}${
      values.username ? `-${values.username}` : ''
    }-${(
      applicationTypes.find(({ id }) => id === app && app.application_type_id) ||
     {
       display_name: `unused-${values.id}`
     }
    ).display_name}`
  }
```

src/components/AddApplication/AddApplicationSchema.js
```diff
@@ -91,10 +198,18 @@ const fields = (applications = [], intl, sourceTypes, applicationTypes, authenti
         }
 
         nextStep = ({ values: { application } }) => {
-            const app = application ? application : {};
-            const appId = app.application_type_id ? app.application_type_id : '';
+            if (application) {
+                const appType = applicationTypes.find(({ id }) => id === application.application_type_id);
```
karelhala, line 202: You are not using `appType` anywhere, right? Might be a good idea to use the function directly
```JS
if (hasAlreadySupportedAuthType(
  authenticationValues,
  applicationTypes.find(({ id }) => id === application.application_type_id),
  sourceType.name
)) {
   return 'selectAuthentication';
}
```

src/components/AddApplication/AddApplicationSchema.js
```diff
@@ -91,10 +198,18 @@ const fields = (applications = [], intl, sourceTypes, applicationTypes, authenti
...
+
+                if (hasAlreadySupportedAuthType(authenticationValues, appType, sourceType.name)) {
+                    return 'selectAuthentication';
+                }
+
+                const app = application ? application : {};
+                const appId = app.application_type_id ? app.application_type_id : '';
 
-            return `${sourceType.name}-${appId}`;
+                return `${sourceType.name}-${appId}`;
```
karelhala, line 211: You can simplify this by directly calculating it
```JS
return `${sourceType.name}-${(application && application.application_type_id) || ''}`;
```

PR #302, opened 2020-04-01
Run check after create app/edit source

rvsia:


rvsia:
where is the travis...?

PR #308, opened 2020-04-02
Remove authentication modal

rvsia:
~~Depends on https://github.com/RedHatInsights/sources-ui/pull/301~~ DONE

Add remove authentication modal

- authentication can be removed only if it is not used by any app

Buttons + list

image

No app attached

image

App is using the authentication

image

Removing state

image

### DIFF
--- a/src/components/SourceEditForm/parser/AuthenticationManagement.js
+++ b/src/components/SourceEditForm/parser/AuthenticationManagement.js
@@ -1,0 +1,57 @@
+import React, { useState } from 'react';
+import { Button } from '@patternfly/react-core/dist/js/components/Button';
+import { FormattedMessage } from 'react-intl';
+import { GridItem } from '@patternfly/react-core/dist/js/layouts/Grid';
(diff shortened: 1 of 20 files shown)


src/components/SourceEditForm/parser/AuthenticationManagement.js
```diff
@@ -0,0 +1,80 @@
...
+                                        id="sources.removeAuthWithApps"
+                                        defaultMessage="used by {appNames}"
+                                        values={{ appNames: appNames.join(', ') }}
+                                    />
+                                    : <FormattedMessage
+                                        id="sources.removeAuthNoApps"
+                                        defaultMessage="not used by any app"
+                                    />
+                            }}
+                        />&nbsp;
```
Hyperkid123, line 61: Is there any reason for this space?

rvsia: not yet :D I had another button there, but the pr is too big, so I removed it and I will add it in the future

PR #326, opened 2020-04-23
Update empty state

rvsia:
- remove card
- make it 64 px from the top

image


Hyperkid123:
@rvsia linter please.

PR #328, opened 2020-04-24
Use tooltip instead of toast/hiding things

### LINKED ISSUE #227: Get proper UX for permissions checker
https://github.com/RedHatInsights/sources-ui/pull/224 introduced permissions checker with a placeholder message.

It needs to get proper UX.

cc @terezanovotna

rvsia:
closes #227

- no toast
- disabled buttons have a tooltip 

permissions

Note: There will be a tooltip in the action menu after https://github.com/patternfly/patternfly-react/issues/4124 is fixed and the fix is released and the version is updated here

### DIFF
--- a/src/pages/Sources.js
+++ b/src/pages/Sources.js
@@ -155,7 +155,7 @@ const SourcesPage = () => {
                             }
                             key="addSourceButton"
                         >
-                            <span>
+                            <span tabIndex="0">
                                 <Button variant='primary' isDisabled id="addSourceButton">
                                     <FormattedMessage
                                         id="sources.addSource"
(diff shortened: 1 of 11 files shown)


src/pages/Sources.js
```diff
@@ -135,15 +136,36 @@ const SourcesPage = () => {
...
+                        <Tooltip
+                            content={
+                                intl.formatMessage({
+                                    id: 'sources.notAdminButton',
+                                    defaultMessage: 'You do not have permission to perform this action.'
+                                })
+                            }
+                            key="addSourceButton"
+                        >
+                            <span>
```
Hyperkid123, line 158: you should add a tab index to the span element. Otherwise, users won't be able to focus it with the keyboard and see the tooltip.

rvsia: fixed :+1:

PR #329, opened 2020-04-28
Hack dropdown item to show tooltips

rvsia:
hack

official solution :man_shrugging:


src/App.scss
```diff
@@ -105,3 +105,8 @@
   // should be 64px from the top, content has padding --pf-global--spacer--lg, so we have to calculate the rest of it
   padding-top: calc(0px - var(--pf-global--spacer--lg) + 64px) !important;
 }
+
+.ins-c-sources__disabled-drodpown-item {
+  pointer-events: auto !important;
+  cursor: default !important;
```
Hyperkid123, line 111: I would keep the disabled cursor.

rvsia: removed!

PR #344, opened 2020-05-14
Update all packages

rvsia:
- updates all packages
- updates tests
- updates wrong submit label in the edit form


Hyperkid123:
@rvsia linter :bomb:

PR #359, opened 2020-06-03
Unify loading/errored steps

rvsia:
Waiting for #358 

Uses loading/errored steps from sources package


src/components/AddApplication/AddApplication.js
```diff
@@ -113,7 +116,16 @@ const AddApplication = () => {
         return  (
             <WizardBody
                 goToSources={goToSources}
-                step={<LoadingStep />}
+                step={<LoadingStep
+                    customText={intl.formatMessage({
```
Hyperkid123, line 120: Why is there the `intl.formatMessage` instead of a formatted message?

rvsia: proptypes.string

I will fix it in the sources package, but after the V2 is fixed

Hyperkid123: Ah i though so.

PR #365, opened 2020-06-09
Small UX fixes

rvsia:
TPINVTRY-961

- removes icons from the table
- change title of review details step in the add application summary
- fix wording when all applications are added
- edit form is disabled on pristine
- reworded remove authentication modal

### DIFF
--- a/src/components/SourceEditForm/SourceEditModal.js
+++ b/src/components/SourceEditForm/SourceEditModal.js
@@ -109,7 +109,7 @@ const SourceEditModal = () => {
                                 isHorizontal: true
                             }}
                             canReset
-                            disableSubmit={['submitting']}
+                            disableSubmit={['submitting', 'pristine']}
                             submitLabel={intl.formatMessage({
                                 id: 'sources.save',
                                 defaultMessage: 'Save'
--- a/src/test/components/sourcesTable/SourcesTable.test.js
+++ b/src/test/components/sourcesTable/SourcesTable.test.js
@@ -119,7 +119,7 @@ describe('SourcesTable', () => {
                 expect(wrapper.find(RowWrapper).first().props().className).toEqual(ROW_WRAPPER_CLASSNAME);
                 expect(wrapper.find(LongArrowAltDownIcon)).toHaveLength(1);
                 expect(wrapper.find(ArrowsAltVIcon)).toHaveLength(expectSortableColumns);
-                expect(wrapper.find('img')).toHaveLength(sourcesDataGraphQl.length);
+                expect(wrapper.find('img')).toHaveLength(0);
                 done();
             });
         });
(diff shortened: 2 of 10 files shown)


src/components/SourceEditForm/SourceEditModal.js
```diff
@@ -109,7 +109,7 @@ const SourceEditModal = () => {
                                 isHorizontal: true
                             }}
                             canReset
-                            disableSubmit={['submitting']}
+                            disableSubmit={['submitting', 'pristine']}
```
Hyperkid123, line 112: is there any async validation? We might want to add `validating` to the list.

src/test/components/sourcesTable/SourcesTable.test.js
```diff
@@ -119,7 +119,7 @@ describe('SourcesTable', () => {
                 expect(wrapper.find(RowWrapper).first().props().className).toEqual(ROW_WRAPPER_CLASSNAME);
                 expect(wrapper.find(LongArrowAltDownIcon)).toHaveLength(1);
                 expect(wrapper.find(ArrowsAltVIcon)).toHaveLength(expectSortableColumns);
-                expect(wrapper.find('img')).toHaveLength(sourcesDataGraphQl.length);
+                expect(wrapper.find('img')).toHaveLength(0);
```
Hyperkid123, line 122: We can just remove this expect.

rvsia: It makes sure that there are no icons :D

Hyperkid123: Well, we could likely check if there are no other elements that are not supposed to be there. We should not check that something does not exist if its never actually there.

src/components/SourceEditForm/SourceEditModal.js
rvsia: Validating is disabling button by default in the common formTemplate

```jsx
disabled={submitting || validating || disableSubmit}
```

src/test/components/sourcesTable/SourcesTable.test.js
rvsia: Before this change the images were there. And I think each change should be tested, otherwise you can't be sure that the change worked. Also, we should test that there are no additional elements on the page, however except useless snapshots, there is no easy way to do it.

Hyperkid123: I disagree with this. Tests are not there to keep history but to test the actual code. There are no images in the code ever. Therefore there is no point to test of there are some.

rvsia: `test the actual code.` It tests the actual code.
`There are no images in the code ever.` Without the test, you can't say it surely.
`expect(true).toBe(true)` is tautology, `expect(wrapper.find('img')).toHaveLength(0)` isn't.

Hyperkid123: > expect(wrapper.find('img')).toHaveLength(0) isn't.

It is. Show me a scenario where there is an img tag in the table.

rvsia: Every scenario, where someone adds the tag in the code. Otherwise, we could throw away testing of every static feature.

PR #368, opened 2020-06-15
Update everything

rvsia:


package.json
```diff
@@ -82,7 +82,7 @@
     "sass-loader": "^8.0.2",
     "source-map-loader": "^1.0.0",
     "style-loader": "^1.2.1",
-    "stylelint": "^13.5.0",
+    "stylelint": "^13.6.0",
```
Hyperkid123, line 85: Now that I see it, is stylelint even used?

rvsia: There is a script for it, but it's not used in CI and I also do not use it.

Hyperkid123: I wonder if we could just get rid of it

rvsia: I tried it and it actually does nothing = it's removed.

PR #370, opened 2020-06-15
Do clean install of package-lock.json

rvsia:
I removed `package-lock.json` to enforce installation of the latest versions (Yeah, I did it [before](https://github.com/RedHatInsights/sources-ui/pull/368) but it had a different result :man_shrugging: )

Should fix
https://bugzilla.redhat.com/show_bug.cgi?id=1846130
https://bugzilla.redhat.com/show_bug.cgi?id=1846449

```
    "serialize-javascript": {
      "version": "3.1.0",
```

```
    "websocket-extensions": {
      "version": "0.1.4",
```

PR #371, opened 2020-06-15
Fix nested dependencies security vulnerabilities

Hyperkid123:
Don't know how but newer version of `webpack` is using an older version of `serialize-javascript` which can have some potential security vulnerabilities. Lovered version of webpack until it gets resolved.

The other one (`websocket-extensions`) was easy, just removed it from the lock file and npm install the latest version.

- https://bugzilla.redhat.com/show_bug.cgi?id=1846138#c0
- https://bugzilla.redhat.com/show_bug.cgi?id=1846449#c0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4331,177 +4331,178 @@
       "dev": true
     },
     "@webassemblyjs/ast": {
-      "version": "1.9.0",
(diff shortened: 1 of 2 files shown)


Hyperkid123:
well, #370 fixes it as well. :shrug:

PR #374, opened 2020-06-29
Fix elliptic vulnerability issue

rvsia:
https://bugzilla.redhat.com/show_bug.cgi?id=1851501


rvsia:
@Hyperkid123 DDF was locked just because of the latest PF4.

PR #384, opened 2020-07-14
Remove NODE_ENV from release script

gmcculloug:


.travis/custom_release.sh
```diff
@@ -2,8 +2,6 @@
 set -e
 set -x
 
-NODE_ENV=production npm run build
```
Hyperkid123, line 5: the `npm run build` bit should stay.

gmcculloug: Are you sure?  😉   Updating.

PR #388, opened 2020-07-20
Update add application steps to the latest design

rvsia:
Update steps for addApplication wizard - same as in https://github.com/RedHatInsights/frontend-components/pull/621

TODO: 
- [x] retry adding
- [x] remove old retry code
- [x] tests

Wait for https://github.com/RedHatInsights/frontend-components/pull/634

**Timeout step**

image

**User input error**

image

Edit variant

addappedit

Remove variant

addappremove


src/api/doAttachApp.js
```diff
@@ -149,7 +156,19 @@ export const doAttachApp = async (
...
         await Promise.all(promisesSecondRound);
+
+        if (applicationDataOut) {
+            const timeout = timeoutedApps(appTypes).includes(applicationDataOut.application_type_id) ? 10000 : 0;
+            return await checkAppAvailability(applicationDataOut.id, timeout);
+        }
+
+        return {};
     } catch (error) {
+        console.error(error);
```
Hyperkid123, line 167: is this console intentional?

rvsia: We are not showing error anywhere in the UI, so I think we should show the error at least in the log.

PR #422, opened 2020-09-17
Update second part of edit

rvsia:
- rebase to master
- groups Authentications to applications

**After**

No app

image

One App

image

Multiple Apps

image


Hyperkid123 (approved):

beccaglass:
@rvsia can we use box tabs instead of open tabs? Also, for the no application example, are these all Cost Management ARNs or can they be used by any application? If Cost, can we change field to 'Cost Management ARN'?

PR #425, opened 2020-09-21
Split unused auths into seperate tab

rvsia:
- unused auths are grouped in a separate tab
- remove auth functionality is reverted

TODO:
- [x] test unused auths part
- [x] add back 'remove auth' functionality 

image


src/components/SourceEditForm/parser/GridLayout.js
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    const isDeleting = source?.authentications?.find(auth => auth.id === id)?.isDeleting;
+
+    if (isDeleting) {
+        return <RemoveAuthPlaceholder />;
+    }
+
+    return (<Grid>
+        <GridItem md={2} className="ins-c-sources__grid-layout">
```
Hyperkid123, line 33: Can you use flex layout instead of grid if you are overriding the display property to flex?

Hyperkid123 (approved):

PR #457, opened 2020-10-14
Do not load cost management values in edit when there is no cost app

rvsia:


src/api/doLoadSourceForEdit.js
```diff
@@ -1,12 +1,12 @@
...
 import { getCmValues } from './getCmValues';
 
-export const doLoadSourceForEdit = (source) =>
+export const doLoadSourceForEdit = (source, hasCostManagement) =>
   Promise.all([
     getSourcesApi().showSource(source.id),
     getSourcesApi().listSourceEndpoints(source.id),
     doLoadApplicationsForEdit(source.id),
-    getCmValues(source.id).catch(() => undefined),
+    hasCostManagement && getCmValues(source.id).catch(() => undefined),
```
Hyperkid123, line 9: This will add `false` to the promise array if `hasCostManagement` is false. Are you sure that is ok? Won't that break the promise resolution?

rvsia: I tested it in the console/ui and it seems ok.

You can double check it by running 

```jsx
Promise.all([new Promise(res => setTimeout(() => res('ok'), 1500)), false]).then((data) => console.log(data));
```

in a console.

Hyperkid123: Well, yea but you will get the `false` as a promise result. That is my point. Can the then branch work with `false` value?

rvsia: Oh yeah, it has been already checking if 'costManagementOutData' is positive, because in most cases it went into catch branch and returned `undefined`.

Hyperkid123 (approved):

PR #482, opened 2020-11-17
Update Dockerfile to use UBI8 image instead of alpine

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10673

Updating the Dockerfile to use ubi8 instead of the alpine image. 

@rvsia - there are 3 different versions of node module streams, I went with the oldest (10) since we were one node 8 before, but could we use one of the newer ones? Here are the options:
```
nodejs                    10 [d]         common [d], development, minimal, s2i     Javascript runtime
nodejs                    12             common [d], development, minimal, s2i     Javascript runtime
nodejs                    14             common [d], development, minimal, s2i     Javascript runtime
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,21 @@
-FROM node:8-alpine
+FROM registry.access.redhat.com/ubi8/ubi:8.2-343
 
 ENV WORKDIR /sources-ui/
 RUN mkdir -p $WORKDIR
 WORKDIR $WORKDIR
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

Hyperkid123:
@lindgrenj6 the v10 is fine for now.

Here you can see a timetable for node versions support: https://nodejs.org/en/about/releases/

PR #487, opened 2020-11-25
Implement drill down page

rvsia:
closes https://issues.redhat.com/browse/RHCLOUD-10840

TODO:
- [x] all actions
  - rename will be done seperately
- check status button
  - will be done seperately
- [x] fix kebab
- [x] tests
- [x] new codecoverage

**After**

No permissions

image

image

With permissions

image


rvsia:
@Hyperkid123 This is ready for review :sweat_smile: 

I know, it seems that it's a lot of new code but actually there are **no new breaking features**, only the layout/routing is changed and all other components are updated to support it:

- All new stuff is in Detail.js, components/SourceDetail/*.js

This is the important stuff, but all these components are pretty basic as you can see.

- App remove modal was updated to fetch data from the URL, not from props
- DataLoader is just a functionality extracted from Sources page
- AddApplication wizard is simplified to generate steps only for the selected app type (via url) so all changes there are just simplification
- From SourceEditRemove everything except applications settings is removed
- most of the changes are tests

If you want, we can go over it together

Hyperkid123 (approved):

PR #490, opened 2020-12-01
Add availability check button to source detail

rvsia:
**After**

check

### DIFF
--- a/src/components/SourceDetail/AvailabilityChecker.js
+++ b/src/components/SourceDetail/AvailabilityChecker.js
@@ -1,0 +1,49 @@
+import React, { useState } from 'react';
+import { useIntl } from 'react-intl';
+import { useDispatch } from 'react-redux';
+
+import RedoIcon from '@patternfly/react-icons/dist/js/icons/redo-icon';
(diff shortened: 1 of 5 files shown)


src/components/SourceDetail/AvailabilityChecker.js
```diff
@@ -0,0 +1,49 @@
...
+            intl.formatMessage({
+              id: 'sources.checkavailability.notificationDescription',
+              defaultMessage: 'Check this page later for updates',
+            })
+          )
+        );
+      }}
+      isDisabled={loading}
+    >
+      {!loading && <RedoIcon />}
```
Hyperkid123, line 43: You should use the `icon` prop if you want icon buttons.

rvsia: No, plain buttons need icons as children. See https://www.patternfly.org/v4/components/button

Hyperkid123: :face_with_head_bandage: consistent API eh?

Hyperkid123 (approved):

PR #494, opened 2020-12-02
Split the table to Cloud sources and RedHat sources

rvsia:
part of https://issues.redhat.com/browse/RHCLOUD-10841

**Before**

image

**After**

image

### DIFF
--- a/src/components/TabNavigation.js
+++ b/src/components/TabNavigation.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { Tabs } from '@patternfly/react-core/dist/js/components/Tabs/Tabs';
+import { Tab } from '@patternfly/react-core/dist/js/components/Tabs/Tab';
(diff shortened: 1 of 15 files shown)


src/components/TabNavigation.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { Tabs } from '@patternfly/react-core/dist/js/components/Tabs/Tabs';
```
Hyperkid123, line 5: Just a note here. It should be safe using esm PF components. They do transform imports from CJS to ESM. We can finally reap all the benefits of ESM build without having duplicate components.

Hyperkid123 (approved):

src/components/TabNavigation.js
rvsia: So should I run find and replace all? :D

Hyperkid123: It "should" :tm: work. You might even don't have to use absolute imports anymore. But I would definitely check the bundle size after that.

PR #529, opened 2021-01-07
Add a command for a run with linked wizard

rvsia:


config/base.webpack.config.js
```diff
@@ -43,7 +43,7 @@ const webpackConfig = {
   },
 
   resolve: {
-    // modules: [resolve('./node_modules')], // uncomment when using linked package
+    ...(process.env.LINKED && { modules: [resolve('./node_modules')] }),
```
Hyperkid123, line 46: Is this suppose to reference the linked package node_modules? Would it be better to resolve the sources node_modules? Like with the aliases bellow?

rvsia: It works, that's enough for me. :man_shrugging:

Hyperkid123 (approved):

PR #560, opened 2021-02-02
Switch sources to use ESM instead of CJS

rvsia:
- [x] https://github.com/RedHatInsights/frontend-components/pull/938
- ~~use esm react table~~

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
-import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/cjs/NotificationPortal';
-import { Main } from '@redhat-cloud-services/frontend-components/components/cjs/Main';
+import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/esm/NotificationPortal';
+import { Main } from '@redhat-cloud-services/frontend-components/components/esm/Main';
 import { IntlProvider } from 'react-intl';
 import { BrowserRouter as Router } from 'react-router-dom';
 
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,0 @@
-{
-    "presets": ["@babel/preset-env", "@babel/preset-react"],
-    "plugins": [
-        "@babel/plugin-transform-runtime",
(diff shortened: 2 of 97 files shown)


src/App.js
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
-import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/cjs/NotificationPortal';
-import { Main } from '@redhat-cloud-services/frontend-components/components/cjs/Main';
+import NotificationsPortal from '@redhat-cloud-services/frontend-components-notifications/esm/NotificationPortal';
+import { Main } from '@redhat-cloud-services/frontend-components/components/esm/Main';
```
Hyperkid123, line 3: You could instead upgrade to v3 FEC dependencies which resolve the asset type automatically.

rvsia: That's exactly what I am planning to do this in follow-up PRs. Now I aiming to switch to ESM, so I can update the wizard package.

src/components/SourcesTable/SourcesTable.js
```diff
@@ -1,7 +1,9 @@
 import React, { useEffect, useReducer } from 'react';
 import { useDispatch, useSelector, shallowEqual } from 'react-redux';
 import { useHistory } from 'react-router-dom';
-import { Table, TableHeader, TableBody, sortable, wrappable } from '@patternfly/react-table';
+
+import { Table, TableHeader, TableBody, sortable, wrappable } from '@patternfly/react-table/dist/esm/index';
```
Hyperkid123, line 5: This is not required. Webpack will use the ESM assets anyway plus the table package does not treeshake react-core anyway :shrug: wont help.

rvsia: Yeah, but I had to use 'cjs' paths in tests for checking table icons. This makes it more transparent.

rvsia: But yeah, you are right, the mixing is not a big price.

Hyperkid123 (approved):

PR #579, opened 2021-02-17
Use DevEntry for dev env.

Hyperkid123:
This way you can have the logger in chrome 2 dev env.


src/DevEntry.js
```diff
@@ -0,0 +1,18 @@
...
+
+const DevEntry = () => (
+  <Provider store={getDevStore()}>
+    <App />
+  </Provider>
+);
+
+DevEntry.propTypes = {
+  store: PropTypes.object,
+};
```
rvsia, line 16: not needed

rvsia (commented):
Can you please also change bootstrap dev so it imports the DevEntry? Thanks!

rvsia (approved):

Hyperkid123:
@rvsia I have added a fix for the faulty redux store. I had to remove the entries tests because the functions are no longer called during the render cycle.

rvsia (approved):

PR #603, opened 2021-03-04
Add AWS account number to summary card

rvsia:
Part of https://issues.redhat.com/browse/RHCLOUD-12147

image

### DIFF
--- a/src/components/SourceDetail/SourceSummaryCard.js
+++ b/src/components/SourceDetail/SourceSummaryCard.js
@@ -31,6 +31,17 @@ const SourceSummaryCard = () => {
   const intl = useIntl();
   const source = useSource();
   const sourceTypes = useSelector(({ sources }) => sources.sourceTypes, shallowEqual);
+  const type = sourceTypes.find((type) => type.id === source.source_type_id);
+
+  let awsAccountNumber;
+
+  if (type.name === 'amazon') {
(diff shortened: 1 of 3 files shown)


src/components/SourceDetail/SourceSummaryCard.js
```diff
@@ -31,6 +31,17 @@ const SourceSummaryCard = () => {
...
   const sourceTypes = useSelector(({ sources }) => sources.sourceTypes, shallowEqual);
+  const type = sourceTypes.find((type) => type.id === source.source_type_id);
+
+  let awsAccountNumber;
+
+  if (type.name === 'amazon') {
+    const arnType = source.authentications?.find(({ authtype }) => authtype.includes('arn'));
+
+    if (arnType?.username) {
+      awsAccountNumber = arnType.username.match(/:\d+/)?.[0].replace(/:/g, '');
```
Hyperkid123, line 42: Maybe we want to add the conditional accessor before the `.replace` as well

rvsia: Hmm, I don't think so, because if there is a match, it will be a string all the time, won't be?

Hyperkid123: Yeah, it should be there. It's just my brain is always warning me about accessing stuff in the array via index :ghost:

Hyperkid123 (approved):

PR #630, opened 2021-03-22
Toolbar dropdown mobile fixes

epwinchell:
- Right align dropdowns in mobile to prevent cropping
- Make buttons emulate link styling when in kebab

Old
Screen Shot 2021-03-22 at 1 47 54 PM

New
Screen Shot 2021-03-22 at 2 18 05 PM

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -178,3 +178,17 @@
   color:	#6a6e73;
   line-height: 24px;
 }
+
+// mobile fixes
+@media only screen and (max-width: 768px) {
+  .pf-c-dropdown__menu { right: 0; } // position all dropdown contents left in mobile
+
+  .sources-toolbar .pf-c-dropdown button { // make button emulate link styling in kebab
+    --pf-c-button--disabled--BackgroundColor: transparent;
+    --pf-c-button--m-primary--BackgroundColor: transparent;
+    --pf-c-button--after--BorderColor: transparent;
+    --pf-c-button--hover--after--BorderWidth: 0;
+    color: var(--pf-global--Color--200);
+  }
+}
+
--- a/src/pages/Sources.js
+++ b/src/pages/Sources.js
@@ -139,6 +139,7 @@ const SourcesPage = () => {
   const mainContent = () => (
     <React.Fragment>
       <PrimaryToolbar
+        className="sources-toolbar"
         pagination={showPaginationLoader ? <PaginationLoader /> : numberOfEntities > 0 ? paginationConfig : undefined}
         actionsConfig={
           hasWritePermissions


src/App.scss
```diff
@@ -178,3 +178,17 @@
   color:	#6a6e73;
   line-height: 24px;
 }
+
+// mobile fixes
+@media only screen and (max-width: 768px) {
+  .pf-c-dropdown__menu { right: 0; } // position all dropdown contents left in mobile
```
rvsia, line 184: We can achieve the same by using `drodpownProps`:

```jsx
        actionsConfig={{
          dropdownProps: { position: 'right' },
          actions: hasWritePermissions
            ? [
                <Link to={routes.sourcesNew.path} key="addSourceButton">
                  <Button variant="primary" id="addSourceButton">
                    {intl.formatMessage({
                      id: 'sources.addSource',
                      defaultMessage: 'Add source',
                    })}
                  </Button>
                </Link>,
              ]
            : [
                <Tooltip
                  content={intl.formatMessage({
                    id: 'sources.notAdminAddButton',
                    defaultMessage:
                      'To add a source, you must be granted write permissions from your Organization Administrator.',
                  })}
                  key="addSourceButton"
                >
                  <span tabIndex="0">
                    <Button variant="primary" isDisabled id="addSourceButton">
                      {intl.formatMessage({
                        id: 'sources.addSource',
                        defaultMessage: 'Add source',
                      })}
                    </Button>
                  </span>
                </Tooltip>,
              ],
        }}
```

src/pages/Sources.js
```diff
@@ -139,6 +139,7 @@ const SourcesPage = () => {
   const mainContent = () => (
     <React.Fragment>
       <PrimaryToolbar
+        className="sources-toolbar"
```
rvsia, line 142: Can you please follow the naming style we are using? 

i.e. `ins-c-source__sources-toolbar`

rvsia:
Can we also make a change to the color of the button? It looks like disabled by default

**Before**

Screenshot 2021-03-23 at 8 53 42

**After**

`color: var(--pf-c-dropdown__menu-item--Color)`

Screenshot 2021-03-23 at 8 54 21

src/App.scss
epwinchell: @rvsia Using drodpownProps doesn't allow us to position it for mobile only.

rvsia: The dropdown is only on mobile, so it's not an issue here.

rvsia:
Closing in favor of https://github.com/RedHatInsights/sources-ui/pull/658

PR #631, opened 2021-03-23
Update dependencies

rvsia:
I also updated tests for SourceDetail page that started failing. I have added 2ms timeout to each and it seems it did the trick, yeah, I tried other options, fake timers, immediate, awaits.. nothing helped. It seems that enzyme has issues with suspense + routes + mocked components. I know the solution is ugly.. but It seems to work and I don't want to spend more time on resolving a stupid issue like this one. However, I welcome any feedback. 😄


Hyperkid123 (approved):
My only feedback would be stop using enzyme and migrate to RTL :wink:

PR #658, opened 2021-04-22
Use dropdown item on mobile screen instead of button

rvsia:
Adds useScreen hook that allows to change components according to size
**Before**

Screenshot 2021-04-22 at 11 50 20

**After**

No permissions

Screenshot 2021-04-22 at 11 37 38

With permissions

Screenshot 2021-04-22 at 11 37 25


Hyperkid123 (approved):

beccaglass:
@rvsia how does the tooltip for the no permissions version work on mobile?

rvsia:
@beccaglass when you click on it it appears

PR #699, opened 2021-06-08
Navigate to sources page on nav click

karelhala:
JIRA: https://issues.redhat.com/browse/RHCLOUD-13096

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,7 +4,7 @@ import { Main } from '@redhat-cloud-services/frontend-components/Main';
 import { IntlProvider } from 'react-intl';
 import { BrowserRouter as Router } from 'react-router-dom';
 
-import Routes from './Routes';
+import Routes, { routes } from './Routes';
 
 import './App.scss';
 import './styles/authSelect.scss';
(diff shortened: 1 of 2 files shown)


karelhala:
ping @rvsia

Hyperkid123:
@rvsia there is a problem that we have two separate history instances. So We will include the fix to the chrome nav re-work, but we need this as a short-term solution.

karelhala:
@rvsia not really, as chrome doesn't have access to the app state and we don't really know where to navigate user in the application. We don't want to issue app refresh as that is not user friendly and we've been sending the nav change for some time, it's just that we are sending the nav update also for active links.

src/App.js
```diff
@@ -28,6 +28,8 @@ const App = () => {
       console.warn('Failed to initialize chrome navigation.');
     }
 
+    insights.chrome.on('APP_NAVIGATION', (event) => event.navId === 'sources' && history.push(routes.sources));
```
rvsia, line 31: please use `routes.sources.path`

rvsia (commented):
I think tests will fail, because insights.on does not exist.

Also, it would be nice to add a test case to check this behavior.

rvsia:
@Hyperkid123 @karelhala
>  we need this as a short-term solution.

so are you going to update all the applications that are currently deployed?

> we don't really know where to navigate user in the application.

I think it's the same as the user clicks for the first time.

rvsia (approved):

PR #701, opened 2021-06-09
Update dependencies

rvsia:
- Update all dependencies
- Replace babel-polyfill with core-js as recommended
- Remove obsolete @babel/plugin-proposal-class-properties


config/setupTests.js
```diff
@@ -1,7 +1,7 @@
 import { configure, mount, render, shallow } from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import React from 'react';
-import 'babel-polyfill';
+import 'core-js';
```
Hyperkid123, line 4: You might not even need `core-js`. Try running the tests without any polyfills.

rvsia: There is still missing `flatMap`, I can remove it after this project is migrated to latest Node.

Hyperkid123 (approved):

PR #702, opened 2021-06-09
Fix navigation event listener

rvsia:
Fix the navigation listener by correctly calling `history.push` and calling it inside the router component.

### DIFF
--- a/src/components/NavigationListener.js
+++ b/src/components/NavigationListener.js
@@ -1,0 +1,23 @@
+import { useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { routes } from '../Routes';
+
+const NavigationListener = () => {
+  const history = useHistory();
+
+  useEffect(() => {
+    insights.chrome.init();
+    try {
+      insights.chrome.identifyApp('sources');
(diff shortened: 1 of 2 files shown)


src/components/NavigationListener.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+
+import { routes } from '../Routes';
+
+const NavigationListener = () => {
```
Hyperkid123, line 6: Could this be just a hook? I don't see any benefit in wrapping it in a react component.

rvsia: There is no component that makes sense to have this functionality included.

Hyperkid123 (approved):

src/components/NavigationListener.js
rvsia: Routes are just routes. 😅 

Creating a wrapper is possible, but still, there would be a component that just does the same, just with children instead of null. Don't see any pros in this approach. 🤔 

Actually, I like this design because it makes the hooks more pluggable without any issues.

PR #719, opened 2021-06-24
Use domEvent.href in the nav listener

rvsia:


Hyperkid123 (approved):

src/components/NavigationListener.js
```diff
@@ -12,7 +12,7 @@ const NavigationListener = () => {
       insights.chrome.identifyApp('sources');
       const unregister = insights.chrome.on(
         'APP_NAVIGATION',
-        (event) => event.navId === 'sources' && history.push(routes.sources.path)
+        (event) => event.domEvent.href.split('/').pop() === 'sources' && history.push(routes.sources.path)
```
Hyperkid123, line 15: It just hit me. The href is only available in `ci-beta` for now. So this will break every other environment. Can you gate it behind `window.insights.chrome.getEnvironment() === 'ci' && window.insights.chrome.isBeta()`

We will be resuming the normal chrome promotion process soon.

rvsia: This gate would be too hard to manage. I commited a code that uses `navId` if it's not an empty string (resp. `/`) and if not, then it uses the `href` attribute.

src/components/NavigationListener.js
```diff
@@ -12,7 +12,9 @@ const NavigationListener = () => {
       insights.chrome.identifyApp('sources');
       const unregister = insights.chrome.on(
         'APP_NAVIGATION',
-        (event) => event.navId === 'sources' && history.push(routes.sources.path)
+        (event) =>
+          (event.navId.split('/').pop() || event.domEvent.href.split('/').pop()) === 'sources' &&
```
Hyperkid123, line 16: The `href` is not even in the old object, you should use `event.domEvent?.href`

Hyperkid123 (approved):

Hyperkid123 (approved):

PR #721, opened 2021-06-29
Add CVS/JSON export button

rvsia:
https://issues.redhat.com/browse/RHCLOUD-14766

**After**

Kapture 2021-06-30 at 09 17 41

TODO:
- [x] tests

### DIFF
--- a/src/pages/Sources.js
+++ b/src/pages/Sources.js
@@ -8,6 +8,8 @@ import { PrimaryToolbar } from '@redhat-cloud-services/frontend-components/Prima
 import { Section } from '@redhat-cloud-services/frontend-components/Section';
 import { ErrorState } from '@redhat-cloud-services/frontend-components/ErrorState';
 
+import { downloadFile } from '@redhat-cloud-services/frontend-components-utilities/helpers';
+
 import { filterSources, pageAndSize } from '../redux/sources/actions';
 import SourcesTable from '../components/SourcesTable/SourcesTable';
 import { routes } from '../Routes';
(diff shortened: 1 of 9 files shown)


Hyperkid123 (approved):

src/pages/Sources.js
```diff
@@ -264,6 +269,16 @@ const SourcesPage = () => {
...
           onDelete: (_event, chips, deleteAll) => dispatch(filterSources(removeChips(chips, filterValue, deleteAll))),
         }}
+        exportConfig={{
+          isDisabled: !loaded,
+          onSelect: (_e, type) => {
+            const data =
+              type === 'csv'
+                ? generateCSV(entities, intl, appTypes, sourceTypes)
+                : generateJSON(entities, intl, appTypes, sourceTypes);
+            downloadFile(data, `sources-${new Date().toISOString()}`, type);
```
Hyperkid123, line 279: As this is very dependent on the API response shape and can cause runtime errors, can we add a try-catch block around the action? And possibly display error notifications if the CSV/JSON conversion fails?

rvsia: If the response is different than the whole app would fail. 🤷

Hyperkid123: Spicy :hot_pepper:

rvsia: There is the error boundary, so it should be ok. 😏 

But the whole application is based on this one API response, so if it were changed, then the whole application would be changed and there are integration tests that would catch all the differences.

PR #753, opened 2021-09-21
Update class prefixes

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-16246

- Replace "ins-" prefix with "src-" and update name formatting
- delete unused classes
- add a few PF variables


rvsia (approved):
Looks great! 🏆

PR #779, opened 2021-11-05
Marketplace recommended services poc

rvsia:
https://issues.redhat.com/browse/RHCLOUD-16841

Kapture 2021-11-10 at 10 57 17

Screenshot 2021-11-09 at 18 05 40


config/dev.webpack.config.js
```diff
@@ -19,7 +20,27 @@ plugins.push(
     exposes: {
       './RootApp': resolve(__dirname, '../src/DevEntry'),
     },
-  })
+  }),
+  new ExtensionsPlugin(
+    {
+      extensions: [
```
karelhala, line 26: We don't want to have route for now? This will render directly in the content, which I guess is fine for now.

src/marketplace/MarketplaceCard.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Card, CardTitle, CardBody, CardFooter, Label, Button } from '@patternfly/react-core';
+import { DatabaseIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
+import ImageWithPlaceholder from '../components/TilesShared/ImageWithPlaceholder';
+
+const MarketplaceCard = (product) => (
+  <Card isFlat className="marketplace-card">
+    <CardTitle className="pf-u-pl-md pf-u-pr-md">
+      <ImageWithPlaceholder src={product.publishedContent.logo.url} style={{ maxHeight: 24 }} />
```
karelhala, line 9: Not sure how the styling is handled, but don't we want to have styles in scss files instead?

src/marketplace/MarketplaceModal.js
```diff
@@ -0,0 +1,109 @@
...
+  page: 1,
+  perPage: 10,
+  itemCount: data?.length || 0,
+  filters: {
+    type: ['database'],
+  },
+});
+
+const reducer = (state, { type, payload }) => {
+  switch (type) {
```
karelhala, line 18: I still prefer object assign over switches, but this is not my repo :smile:

karelhala (approved):
From the extensions perspective everything looks good! Just a few questions, but good to merge.

config/dev.webpack.config.js
rvsia: I wasn't sure what we need here.

karelhala: If you want to have just a route in the navigation you can add, it should just show Marketplace in the navigation
```
{
      "type": "console.navigation/href",
      "properties": {
        "id": "dev-dynamic-route-1",
        "name": "Marketplace",
        "href": "/marketplace"
      }
}
```

src/marketplace/MarketplaceCard.js
```diff
@@ -0,0 +1,37 @@
...
+      {product.publishedContent.localizedContent.descriptions.short}
+    </CardBody>
+    <CardFooter className="pf-u-pl-md pf-u-pr-md pf-u-pb-md">
+      <Button
+        variant="link"
+        isInline
+        icon={<ExternalLinkAltIcon />}
+        iconPosition="right"
+        component="a"
+        href={`https://sandbox.marketplace.redhat.com/en-us/products/${product.productPageName}`}
```
karelhala, line 27: You should use the `MARKETPLACE_URL` in here as well.

karelhala (approved):

config/dev.webpack.config.js
rvsia: I thought we were talking about just having a direct access via url. 🤔 

But until the gateway is ready, we can let it as it is. There will be more changes otherwise, this would be fine for Demo purposes.

karelhala: Yeah, good point!

src/marketplace/api.js
```diff
@@ -0,0 +1,58 @@
...
+    apikey: localStorage.getItem('marketplace-key'),
+  });
+
+const query = `
+{
+  productByIds (ids: [
+    "ca54ff2ba24df884a96995f1eefd081b",
+    "ccb29628eabe15c81121ecf231a0a8cd",
+    "9e3b238934c288394680f3a53d1544e0",
+    "bd44b472123b544a630a939c650e968e"
```
Hyperkid123, line 16: Is it possible to get these ids from API? Or do we have the static list for now? I am just thinking about other envs.

src/marketplace/api.js
```diff
@@ -0,0 +1,58 @@
...
+        content { name }
+      }
+      company {
+        displayName
+      }
+    }
+  }
+}`;
+
+let access_token;
```
Hyperkid123, line 45: For how long is the token valid? This may cause issues once the token expires.

src/marketplace/api.js
rvsia: This is just a POC, it will be definitely changed in the future. For now, we just show few selected services.

src/marketplace/api.js
rvsia: I wasn't actually able to expired it out, it is valid for a long time. However, as it is just a POC and the external api will be replaced with an internal service soon, I wouldn't care too much.

Hyperkid123 (approved):

PR #804, opened 2022-01-28
[RHCLOUD-17554] Add IBM type to wizard

rvsia:
Fixes: https://issues.redhat.com/browse/RHCLOUD-17554

TODO:
- [x] tests (render description, wizard flow)
- [ ] update API backend https://github.com/RedHatInsights/sources-api/pull/476

**After**

Kapture 2022-01-28 at 13 25 58
Screenshot 2022-01-28 at 13 26 02
Screenshot 2022-01-28 at 13 26 06
Screenshot 2022-01-28 at 13 26 10
Screenshot 2022-01-28 at 13 26 14

⚠️  **I had to change formatting to CODE, as the original PF component does not support multiline values. Should i keep this way, should we change it to multiple clipboard components or should we implement a new component similar to that component in mockups (a lot of effort)?** ⚠️ 
Screenshot 2022-01-28 at 14 50 20

Screenshot 2022-01-28 at 13 26 29
Screenshot 2022-01-28 at 13 26 34

**UPDATED API KEY SCREEN**

Screenshot 2022-01-28 at 16 21 06
Screenshot 2022-01-28 at 16 21 37

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -1,4 +1,4 @@
-import { configure, mount, render, shallow } from 'enzyme';
+import { configure, mount, shallow } from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import React from 'react';
 import '@testing-library/jest-dom/extend-expect';
@@ -8,7 +8,6 @@ import 'whatwg-fetch'; // fetch for Nodejs
 configure({ adapter: new Adapter() });
 
 global.shallow = shallow;
-global.render = render;
 global.mount = mount;
 global.React = React;
 
--- a/src/components/FormComponents/SourceWizardSummary.js
+++ b/src/components/FormComponents/SourceWizardSummary.js
@@ -165,7 +165,7 @@ const SourceWizardSummary = ({ sourceTypes, applicationTypes, showApp, showAuthT
 
   const skipEndpoint = shouldSkipEndpoint(type.name, hasAuthentication, name);
 
-  let endpointFields = type.schema.endpoint.fields;
+  let endpointFields = type.schema.endpoint?.fields || [];
 
   if (skipEndpoint) {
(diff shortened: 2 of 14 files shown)


rvsia:
@nlcwong can we get a UX review here? 🙏

nlcwong:
> ⚠️ **I had to change formatting to CODE, as the original PF component does not support multiline values. Should i keep this way, should we change it to multiple clipboard components or should we implement a new component similar to that component in mockups (a lot of effort)?** ⚠️ Screenshot 2022-01-28 at 14 50 20

This is where I'll need to rely on @chambridge to let us know whether or not IBM Cloud Shell allows us to run multiple commands in one go (if we copy and paste the chunk above into the shell), or if we need the user to copy and paste each line into the shell in order for this all to work ... OR if we we can somehow collapse those 4 commands into 1 command? 

@rvsia in terms of style, I rather have the new component where the style is similar to the mocks just because its legible (in my eyes), how about the reverse where if you need to expand out the textbox, then it looks like code (even if it isn't multi-line)?

Also, is there a way to check to make sure that whatever the user entered for the Service ID, that we make sure it does not contain a space in it and if it does, encapsulate the string with quotes so the command won't fail b/c there was a space in the service ID?  I don't think IBM would ever create a service ID that has space in it, but who knows?!

rvsia:
@nlcwong 

1) I wrapped the service id in quotes by default.

2) So I made a custom component:

Screenshot 2022-02-10 at 9 15 43

what do you think?

chambridge:
Tested this out this morning the IBM Cloud Shell treats each newline like an enter so it will essentially execute the first 3 command when they are pasted into the shell and the user can click enter to execute the final line. So copying the block of commands does work here.

Hyperkid123 (approved):

PR #831, opened 2022-02-24
Expose marketplace widget in fed modules

rvsia:


config/prod.webpack.config.js
```diff
@@ -10,6 +10,9 @@ const { config: webpackConfig, plugins } = config({
 plugins.push(
   require('@redhat-cloud-services/frontend-components-config/federated-modules')({
     root: resolve(__dirname, '../'),
+    exposes: {
```
Hyperkid123, line 13: When configuring the expose object, you also have to expose the `./RootApp` module. The default config is overriden.

Hyperkid123 (changes requested):

config/prod.webpack.config.js
rvsia: Fixed

Hyperkid123 (approved):

PR #854, opened 2022-03-25
Convert more tests to RTL

rvsia:


src/components/TilesShared/ImageWithPlaceholder.js
```diff
@@ -9,7 +9,13 @@ const ImageWithPlaceholder = ({ src, ...rest }) => {
   return (
     <React.Fragment>
       {!loaded && <Loader height="40px" width="110px" {...rest} />}
-      <img src={src} onLoad={() => setLoaded(true)} style={{ display: loaded ? 'initial' : 'none' }} {...rest} />
+      <img
+        src={src}
+        data-testid="ImageWithPlaceholder"
```
karelhala, line 14: I guess it should be better to use `alt` and later use `ByAltText`.

src/test/components/credentialsForm/credentialsForm.test.js
```diff
@@ -55,51 +50,35 @@ describe('CredentialsForm', () => {
...
+    expect(
+      screen.getByText(
+        'Use the fields below to reset your account authorization credentials. It may take some time to validate new information.'
+      )
+    ).toBeInTheDocument();
+
+    expect([...screen.getAllByRole('textbox')].map((e) => e.name || e.getAttribute('aria-label'))).toEqual([
+      'authentication.username',
+      'pre-filled-authentication',
+    ]);
```
karelhala, line 81: You could use `getByRole` and setting `name` option. Also these should be screen readable I don't think dotted and hyphened text are of any help.

src/test/components/credentialsForm/credentialsForm.test.js
```diff
@@ -116,71 +95,49 @@ describe('CredentialsForm', () => {
...
-    });
-    wrapper.update();
+    await waitFor(() => expect(() => screen.getByRole('progressbar')).toThrow());
 
-    expect(wrapper.find(MemoryRouter).instance().history.location.pathname).toEqual(
-      replaceRouteId(routes.sourcesDetail.path, sourceId)
-    );
+    userEvent.click(screen.getByText('Cancel'));
+
+    expect(screen.getByTestId('location-display').textContent).toEqual(replaceRouteId(routes.sourcesDetail.path, sourceId));
```
karelhala, line 140: This can be bypassed by calling
```
screen.getByRole('WHAT_IS_THE_ELEMENT?', { name: replaceRouteId(routes.sourcesDetail.path, sourceId) })
```

src/test/components/credentialsForm/credentialsForm.test.js
```diff
@@ -193,57 +150,38 @@ describe('CredentialsForm', () => {
...
-    await act(async () => {
-      wrapper.find('form').simulate('submit');
-    });
-    wrapper.update();
+    userEvent.click(screen.getByText('Submit'));
 
-    expect(wrapper.find(MemoryRouter).instance().history.location.pathname).toEqual(
-      replaceRouteId(routes.sourcesDetail.path, sourceId)
-    );
+    expect(screen.getByTestId('location-display').textContent).toEqual(replaceRouteId(routes.sourcesDetail.path, sourceId));
```
karelhala, line 170: Same as above, no need to check for the `textContent` you can use directly option's `name` attribute.

src/test/components/credentialsForm/credentialsForm.test.js
```diff
@@ -252,51 +190,33 @@ describe('CredentialsForm', () => {
...
 
-    await act(async () => {
-      wrapper.find('form').simulate('submit');
-    });
-    wrapper.update();
-
-    expect(wrapper.find(MemoryRouter).instance().history.location.pathname).toEqual(
-      replaceRouteId(routes.sourcesDetail.path, sourceId)
-    );
+    expect(screen.getByTestId('location-display').textContent).toEqual(replaceRouteId(routes.sourcesDetail.path, sourceId));
```
karelhala, line 209: And here as well.

src/test/components/customRoute/CustomRoute.test.js
```diff
@@ -55,25 +41,21 @@ describe('CustomRoute', () => {
...
       ...route,
       writeAccess: true,
     };
 
-    const wrapper = mount(componentWrapperIntl(<CustomRoute exact route={route} Component={PokusComponent} />, store, ['/path']));
+    render(componentWrapperIntl(<CustomRoute exact route={route} Component={PokusComponent} />, store, ['/path']));
 
-    expect(wrapper.find(PokusComponent)).toHaveLength(1);
-    expect(wrapper.find(RedirectNoWriteAccess)).toHaveLength(1);
+    expect(screen.getByTestId('location-display').textContent).toEqual('/sources');
```
karelhala, line 58: It could be used here as well.

karelhala (commented):
Just a few improved selectors in tests and one addition of `alt` to image and we are golden!

src/test/components/credentialsForm/credentialsForm.test.js
rvsia: The current url is rendered as a child of the element. Location-display is a helper to get the current location displayed.

src/components/TilesShared/ImageWithPlaceholder.js
rvsia: This is purely just for the test of the component itself. In other components, aria labels are passed with src. (ImageWithPlaceholder does not hold any content itself so adding an alt "Image" or something like that would not make much sense)

karelhala: Went trough the usage and looks like every place passes down `alt` so we are good. We might want to consider adding `alt` as required prop. But I'll keep that on you.

src/test/components/credentialsForm/credentialsForm.test.js
rvsia: name is a form attribute, not sure how to make it screen readable. There are `<labels>` for that.

rvsia: However, using `name` would cause issues if there is a change: it will throw an error if it is changed.

This approach using `getAll` and `toEqual` will show a new results after a change (expected vs received), so it is far more developer friendly.

rvsia: I changed the aria label to be human readable. However, I guess the whole application would require some "blind" test to get everything correctly.

karelhala: Yup, this could work
```JS
screen.getByRole('textbox', { name: 'Access key ID' });
screen.getByRole('textbox', { name: 'Filled password' });
```

karelhala: Yeah, if there are any new elements in the form we won't break the test. Isn't there a selector that would fail if found anything else? I would prefer the `name` selector because it forces the devs to make the form screen readers compliant.

rvsia: Yeah, this would work almost the same just a little bit worse because it would throw errors in case of errors. I would rather want to see the changed value as with my previous approach. Or if you add a third element, this test will pass, etc.

karelhala: We can keep as is, it's already screen readable and has all the goodies. I just wanted to have it exposed. But you are correct on the need of updating test when label for `authentication.username` changes.

karelhala (approved):

PR #855, opened 2022-03-29
Add deploy scripts

rvsia:
cc @esebesto

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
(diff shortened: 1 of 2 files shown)


karelhala (approved):
Looking good! Is your app onboarded to app-sre yet? If so you might want to add the deploy config[1] as well, it will auto-update the image.

[1] https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/deploy/frontend.yaml

rvsia:
@karelhala According to docs `Once your container is built and pushed to quay, you can create a CRD for it in a new deploy folder.` that should be done after pushing to quay.

PR #864, opened 2022-04-07
Update RTL User Event to v14

rvsia:
https://github.com/testing-library/user-event/releases/tag/v14.0.0

- created delayed mockApi to mock timeouted apis request (should be replaced with a proper way, see [here](https://github.com/testing-library/user-event/discussions/911))
- `userEvent` should use `init()` functions

- [x] tooltip test - how to trigger a PF tooltip?

I haven't found a way to trigger the PF tooltip using the async userEvent - seems that there is some delay that together with async break it. So I wrote a simple mock of the tooltip.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -32,6 +32,6 @@ global.innerWidth = 1080;
 
 Element.prototype.scrollTo = () => {};
 
-global.mockApi = (data, timeout = 60) =>
+global.mockApi = (data, timeout = 200) =>
   jest.fn().mockImplementation(() => new Promise((resolve) => setTimeout(() => resolve(data), timeout)));
-global.mockApiError = (data, timeout = 60) => jest.fn(() => new Promise((_, reject) => setTimeout(() => reject(data), timeout)));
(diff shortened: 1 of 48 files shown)


config/setupTests.js
```diff
@@ -32,6 +32,6 @@ global.innerWidth = 1080;
 
 Element.prototype.scrollTo = () => {};
 
-global.mockApi = (data, timeout = 60) =>
+global.mockApi = (data, timeout = 200) =>
   jest.fn().mockImplementation(() => new Promise((resolve) => setTimeout(() => resolve(data), timeout)));
-global.mockApiError = (data, timeout = 60) => jest.fn(() => new Promise((_, reject) => setTimeout(() => reject(data), timeout)));
+global.mockApiError = (data, timeout = 200) => jest.fn(() => new Promise((_, reject) => setTimeout(() => reject(data), timeout)));
```
Hyperkid123, line 37: Any specific reason for increasing the timout time? It will slow down the test run.

rvsia: This has to be long enough to not be too fast in slower env.

It is just a temporary solution, in the next step I plan to refactor it so the promise would be triggered manually (no timeout) but that would be done separately. See https://issues.redhat.com/browse/RHCLOUD-18835

PR #936, opened 2022-08-04
Update dependabot.yml

rvsia:
Remove myself from dependabot


lpichler (approved):

Hyperkid123 (changes requested):
@rvsia I don't want these alerts :D

But srsly can you use the `RedHatInsights/platform-experience` team tag instead?

Hyperkid123 (approved):

PR #940, opened 2022-08-17
Merge master -> stable

lindgrenj6:
**_This one looks a bit different...are we sure this branch is used anywhere?_**


Hyperkid123:
@lindgrenj6 by default it should be used in stage-stable. But I see in your custom release that stage-stable is updated from master branch: https://github.com/RedHatInsights/sources-ui/blob/master/.travis/custom_release.sh#L5

So no you can close this PR and delete the branch.

PR #950, opened 2022-09-13
fix(addApplication): enable missing endpoints schema partial

Hyperkid123:
fixes: https://issues.redhat.com/browse/RHCLOUD-21148


lindgrenj6 (approved):
:tada:

PR #954, opened 2022-09-20
Add lighthouse steps for Azure and RHEL

fhlavac:
This adds back reverted https://github.com/RedHatInsights/sources-ui/pull/919

https://issues.redhat.com/browse/RHCLOUD-18688

### DIFF
--- a/src/test/__mocks__/@unleash/proxy-client-react.js
+++ b/src/test/__mocks__/@unleash/proxy-client-react.js
@@ -1,0 +1,8 @@
+const lib = jest.requireActual('@unleash/proxy-client-react');
+
+const unleash = {
+  ...lib,
+  useFlag: jest.fn(() => false),
+};
+
+module.exports = unleash;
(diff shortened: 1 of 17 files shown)


Hyperkid123:
@lindgrenj6 are we positive we can now add back the steps? Is not going to create the same issues as it did the first time?

lindgrenj6:
We _only_ want this in the stage environment so the app team can test it.

So basically until this is prod-ready we'll have to cherry-pick commits to promote them to prod, luckily not a lot of changes happen!

Hyperkid123:
@fhlavac can you add a condition to only enable this in stage? We don't want to block other deployments. Possibly use feature flags and load them during the init process.

src/test/__mocks__/@unleash/proxy-client-react.js
```diff
@@ -0,0 +1,8 @@
+const lib = jest.requireActual('@unleash/proxy-client-react');
+
+const pf = {
+  ...lib,
+  useFlag: jest.fn(() => false),
+};
+
+module.exports = pf;
```
Hyperkid123, line 8: This is not PF :wink:

Hyperkid123 (commented):
Just a little nitpick

Hyperkid123 (approved):

PR #972, opened 2022-10-18
Show "Waiting for update" when check in progress

fhlavac:
This should address https://issues.redhat.com/browse/RHCLOUD-21154

Replaced "Not checked yet" message with "Waiting for update" by default


lindgrenj6:
So for this one - I think we want to just change the text from "not checked yet" to "waiting for update" when they press the button, that way the UI has some feedback rather than no text changing. 

The end-users don't quite understand it's an async process, so when the UI doesn't change they don't think anything is happening.

fhlavac:
@lindgrenj6 yeah, that's what we should be doing now - when there is no `source.last_checked_at` and no `source.last_available_at` timestamp, we display the "waiting for update" instead of "not checked yet"

epwinchell (approved):

lindgrenj6:
Yes - I like that change, the other part of it though is like physically changing the text to "waiting for update" after clicking.

like this:

https://user-images.githubusercontent.com/5856504/198078996-571e8467-3f8d-4fe1-b509-07359d01f2e2.mp4

fhlavac:
@lindgrenj6 sure thing, I'll add it

lindgrenj6:
Yes, that's perfectly fine. The initial issue was the the user was confused and thought the button did nothing even though it very much did.

fhlavac:
@lindgrenj6 Ok I'll add it in a bit!

fhlavac:
@lindgrenj6 after a click, the user will now see "Waiting for update" until he refreshes the page. The status pending flag is stored in the application state

Hyperkid123 (approved):

PR #976, opened 2022-10-21
Remove initial availability check from UI on create

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-14015

Removed availability check on creating source with `ACCOUNT_AUTHORIZATION` creation workflow

### DIFF
--- a/src/api/createSuperSource.js
+++ b/src/api/createSuperSource.js
@@ -1,5 +1,4 @@
 import { getSourcesApi } from './entities';
-import checkSourceStatus from './checkSourceStatus';
 import { checkAppAvailability } from './getApplicationStatus';
 import handleError from './handleError';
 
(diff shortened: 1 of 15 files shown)


lindgrenj6:
I pulled this locally and it looks like it's still calling check_availability on both flows (superkey/manual). 

I also noticed when it's "checking the status" it's just polling `/applications/:id` over and over - we could skip that step for superkey sources since they won't go available for 30+ seconds most of the time.

fhlavac:
@lindgrenj6 thank you for taking a look. It should now behave like this for superkey.
chrome-capture-2022-9-24 (1)

karelhala (approved):

lindgrenj6 (approved):
Thank you!

PR #1007, opened 2022-12-09
Enable HCS for AWS

fhlavac:
For now, removed links to docs until HCS is GA

https://issues.redhat.com/browse/RHCLOUD-20983

Manual mode:
[screen-capture (1).webm](https://user-images.githubusercontent.com/50696716/224315093-27fb1413-dfde-4ae5-a8ca-a4adca3bcf33.webm)

Account auth.:
[screen-capture (2).webm](https://user-images.githubusercontent.com/50696716/224315277-03337321-b9c5-4a6b-a735-092fbe76988f.webm)

### DIFF
--- a/src/components/FormComponents/SourceWizardSummary.js
+++ b/src/components/FormComponents/SourceWizardSummary.js
@@ -18,7 +18,7 @@ import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import ValuePopover from './ValuePopover';
 import hardcodedSchemas from '../../components/addSourceWizard/hardcodedSchemas';
 import { ACCOUNT_AUTHORIZATION } from '../constants';
-import { CLOUD_METER_APP_NAME, COST_MANAGEMENT_APP_NAME } from '../../utilities/constants';
+import { CLOUD_METER_APP_NAME, COST_MANAGEMENT_APP_ID, COST_MANAGEMENT_APP_NAME, HCS_APP_NAME } from '../../utilities/constants';
 import { NO_APPLICATION_VALUE } from '../../components/addSourceWizard/stringConstants';
 import {
   getAdditionalSteps,
@@ -28,8 +28,8 @@ import {
 } from '../../components/addSourceWizard/schemaBuilder';
 import { useFlag } from '@unleash/proxy-client-react';
 
-const alertMapper = (appName, sourceType, intl) => {
-  if (appName === COST_MANAGEMENT_APP_NAME && sourceType !== 'google') {
+const alertMapper = (appName, sourceType, intl, hcsEnrolled) => {
+  if (appName === COST_MANAGEMENT_APP_NAME && sourceType !== 'google' && !hcsEnrolled) {
     return (
       <Alert
         variant="info"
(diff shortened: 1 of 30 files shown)


Hyperkid123:
@fhlavac can add the .cache folder to gitignore?

nlcwong:
My fault! I didn't catch it in my mocks until I saw the implementation, but on Step 4, where the user is selecting the HCS bundle, I realized that my two mocks (unselected HCS & selected HCS) had two different explanation strings and you had used the first mock which had the old string version.
Can you replace:
- "Track Red Hat spend with regardless of point of purchase" with "Track Red Hat committed spend"
- "Unlock cloud images in Microsoft Azure and bring your own subscription instead of paying hourly" with "Track spend through <cloud provider type> and apply them to your Red Hat committed spend." where <cloud provider type> would be AWS, Microsoft Azure, or Google Cloud Platform, depending on what they chose in Step 1?

nlcwong:
Also, i just wanted to make this clear that we should ONLY see the HCS bundle if the customer is participating in the HCS program (which I believe there is a flag to check for it), otherwise, it'll default to the regular cost management flow that we have today, correct?

fhlavac:
> Also, i just wanted to make this clear that we should ONLY see the HCS bundle if the customer is participating in the HCS program (which I believe there is a flag to check for it), otherwise, it'll default to the regular cost management flow that we have today, correct?

Yeah, that's how it works. I'll update the strings. Thank you 🙂

src/components/FormComponents/SourceWizardSummary.js
```diff
@@ -28,8 +28,8 @@ import {
 } from '../../components/addSourceWizard/schemaBuilder';
 import { useFlag } from '@unleash/proxy-client-react';
 
-const alertMapper = (appName, sourceType, intl) => {
-  if (appName === COST_MANAGEMENT_APP_NAME && sourceType !== 'google') {
+const alertMapper = (appName, sourceType, intl, hcsEnrolled) => {
```
Hyperkid123, line 31: Can this be a regular React component? The argument list is starting to grow and its being used as a ReactNode

src/components/addSourceWizard/descriptions/HybridCommittedSpendDescription.js
```diff
@@ -0,0 +1,53 @@
...
+import Point from './Point';
+import { ACCOUNT_AUTHORIZATION } from '../../constants';
+
+const HybridCommittedSpendDescription = ({ id }) => {
+  const intl = useIntl();
+  const { getState } = useFormApi();
+
+  const values = getState().values;
+
+  const isEnabled =
```
Hyperkid123, line 16: Can you wrap this variable init in a `useMemo` hook? We don't want to run the same condition on each render. I don't expect the ID to change very often.

src/components/addSourceWizard/hardcodedComponents/aws/arn.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import PropTypes, { bool } from 'prop-types';
```
Hyperkid123, line 2: Probably unintentional named import?

nlcwong:
@fhlavac Use the names you have in your source type definition.

fhlavac:
image
updated strings & resolved comments

cc @Hyperkid123  @nlcwong

Hyperkid123 (approved):
LGTM

PR #1031, opened 2023-02-15
RHCLOUD-22885: Update Azure source creation with cost export scope capabilities

chambridge:
*  Add the ability to collect cost export scope and export name
* Reorder some of the flows

Related to: https://issues.redhat.com/browse/RHCLOUD-22885


chambridge:
New and reordered steps for the Cost Management flow for Azure:
image

Next collect the cost export scope:
image
Entering the value is optional for backwards compatibility with subscription level only cost exports.

Collect the resource group and storage account info:
image

Obtain the cost export name:
image
This is also optional for backwards compatibility.

Gather the subscription guid for the storage account where the cost export is being stored:
image

Gather credentials and present role configuration steps with scope added:
image

Review steps:
image

nlcwong:
1. Wizard steps doesn't add up.  The select application step wouldn't happen until you get to Step 3 b/c the first step is to select a source type (Azure in this case), then Step 2 is enter in a name of the source and then step 3 would be to select an application
2. "Step 3" We need to re-word things there.
- The intro text needs to be a bit more forthcoming as to what scopes will help with their needs, so putting a blurb that suggests a default scope to use might help. Like "Red Hat recommends you use the billing id as the scope because it is awesome". 
- We need to be more explicit that the user needs to pick one of the commands (not all) and removing the bullets can help with that as well as wording and text hierarchy. 
- Will those commands output the exact string that would need to be entered into the cost export scope input field?  If not, can we devise a simple way of the user entering the id only, but we generate the right prefix for them to enter the id?
3.  "Step 5" Where does the user get the cost export name? Is that something they enter during the creation of the export?  If so, we need to be explicit as to where that name comes from. so above the input field's label, just a sentence of "Enter the name of the cost export that you just created here:" or something like that.
4. "Step 6" if the user enters the subscription id in "step 3", is it possible we can auto-populate this step? (thinking of ways to just cut down the amount of copy/paste the user would have to do)

nlcwong:
@chambridge Here are the mocks with some notes on what was discussed: https://www.sketch.com/s/c21e16ea-f8e6-41f6-8688-da2aa0db7a41/a/4aRvQyA

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +208,77 @@ export const ExportSchedule = () => {
...
+          defaultMessage: 'From the Azure portal, select the scope for the new cost export.',
+        })}
+      </Text>
+      <TextContent className="list-align-left">
+        <TextList component={TextListVariants.ul}>
+          <TextListItem component={TextListItemVariants.li}>
+            {intl.formatMessage({
+              id: 'cost.azure.scopeDescription',
+              defaultMessage:
+                'For a cost export for a billing profile, provide the "id" for the billing profile from the command below.',
```
ragevou, line 250: this step is confusing. Does it mean that I need to provide an id of a billing profile for a cost export for that specific billing profile?

src/test/addSourceWizard/addSourceWizard/hardCodedComponents/cost_management_azure.test.js
```diff
@@ -27,7 +27,9 @@ describe('Cost Management Azure steps components', () => {
     render(<Cm.SubscriptionID />);
 
     expect(
-      screen.getByText('Run the following command in Cloud Shell to obtain your Subscription ID and enter it below:')
+      screen.getByText(
+        'Run the following command in Cloud Shell to obtain the Subscription ID where the cost export is being stored and enter it below:'
```
ragevou, line 31: could it be split into two different steps:
* Run the following command in a Cloud Shell to get the Subscription ID and copy the result.
* Paste the result number in the *Subscription ID* field below.

src/test/addSourceWizard/addSourceWizard/hardCodedComponents/cost_management_azure.test.js
```diff
@@ -97,15 +142,38 @@ describe('Cost Management Azure steps components', () => {
       )
     ).toBeInTheDocument();
     expect(screen.getByText('Learn more')).toBeInTheDocument();
-    expect(screen.getByText('From the Azure portal, add a new export.')).toBeInTheDocument();
+    expect(screen.getByText('From the Azure portal, add a new cost export.')).toBeInTheDocument();
     expect(
       screen.getByText(
         'Provide a name for the container and directory path, and specify the below settings to create the daily export. Leave all other options as the default.'
```
ragevou, line 148: directory path where the container is or which directory?

chambridge:
@nlcwong I have made adjustments based on the mocks:
For example collecting the cost export scope:
image

image

image

User is now directed to enter in the input here:
image

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -112,10 +118,15 @@ export const ConfigureRolesDescription = () => {
 const InternalReaderRoleDescription = () => {
   const form = useFormApi();
   const {
-    values: { authentication },
+    values: { authentication, application },
   } = form.getState();
   const intl = useIntl();
 
+  let scope = '/subscriptions/' + application?.extra?.subscription_id;
```
fhlavac, line 125:
```suggestion
  let scope = application?.extra?.scope || `subscriptions/${application?.extra?.subscription_id}`;
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +213,280 @@ export const ExportSchedule = () => {
...
+        {intl.formatMessage({
+          id: 'cost.azure.exportNameFollowup',
+          defaultMessage: 'After configuring the daily export, enter the following:',
+        })}
+      </Text>
     </TextContent>
   );
 };
+
+class CostExportSelect extends React.Component {
```
fhlavac, line 245: Can we use functional component instead of a class?

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +213,280 @@ export const ExportSchedule = () => {
...
+      </Select>
+    );
+  }
+}
+
+CostExportSelect.propTypes = {
+  handleSelect: PropTypes.func.isRequired,
+};
+
+function CostExportDescription({ selection }) {
```
fhlavac, line 309:
```suggestion
const CostExportDescription = ({ selection }) => {
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +213,280 @@ export const ExportSchedule = () => {
...
+        <Text component={TextVariants.p}>
+          {intl.formatMessage({
+            id: 'cost.azure.invoiceSectionInput',
+            defaultMessage: 'After running the command, enter the output in the following field:',
+          })}
+        </Text>
+      </TextContent>
+    ),
+  };
+  if (selection) {
```
fhlavac, line 453:
```suggestion
  return selection ? selectionText[selection] : selectionText.Subscription;
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +213,280 @@ export const ExportSchedule = () => {
...
+    return selectionText[selection];
+  } else {
+    return selectionText.Subscription;
+  }
+}
+
+export const ExportScope = () => {
+  const intl = useIntl();
+  const [selection, setSelection] = useState('');
+  let selected = 'Subscription';
```
fhlavac, line 463: this could be probably simplified
```
const handleSelect = (selection = 'Subscription') => {
    setSelection(selected);
};
```

fhlavac:
Looks good, just some nitpicks :slightly_smiling_face:

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: updated

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: updated. thanks

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: updated

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: Yep. that's what you get from a backend dev 😆

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: updated

nlcwong:
@chambridge remind me again, are we mandating that they make scope a required step?  I think we said we could going forward and just auto populate the old ones to just say subscription scope, correct?  If so, can we make sure all the input fields are marked with a "Required" asterik, that should disable the "Next" button until the user inputs something into those input fields.  This is also including the export name step as well (that should be required ... no?)

chambridge:
@nlcwong 
No scope is not a required step, as it would mean we wouldn't be backwards compatible. The preference is that new sources will use this though.

chambridge:
Ok. Documentation MR has been reviewed and is ready to go so I'm moving this to ready for review so we can coordinate how this lands in stage & production.

Hyperkid123:
@chambridge we just need dates when to merge/promote this PR.

chambridge:
@Hyperkid123 I think we are good for this to go into stage. Would like to kick the end-to-end flow in stage and confirm with Cost QE, but would think we could promote to production next week if we don't find any gotchas. I can reach out on Slack once Stage looks good.

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +210,258 @@ export const ExportSchedule = () => {
...
+      <SelectOption key="invoiceSection" value="Invoice Section" />
+    </Select>
+  );
+};
+
+CostExportSelect.propTypes = {
+  handleSelect: PropTypes.func.isRequired,
+};
+
+export default CostExportSelect;
```
fhlavac, line 288: this default export does not seem to be needed, also it should be at the bottom of the file

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +210,258 @@ export const ExportSchedule = () => {
...
+        })}
+      </Text>
     </TextContent>
   );
 };
+
+const CostExportSelect = ({ handleSelect }) => {
+  const [isOpen, setOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState('Subscription');
+  const onToggle = (isOpen) => {
```
fhlavac, line 245:
```suggestion
  const onToggle = (isOpen) => setOpen(isOpen);
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -202,10 +210,258 @@ export const ExportSchedule = () => {
...
+      </TextContent>
+    ),
+  };
+  return selection ? selectionText[selection] : selectionText.Subscription;
+};
+
+export const ExportScope = () => {
+  const intl = useIntl();
+  const [selection, setSelection] = useState('');
+  const handleSelect = (selection = 'Subscription') => {
```
fhlavac, line 440:
```suggestion
  const handleSelect = (selection = 'Subscription') => setSelection(selection);
```

fhlavac:
Just one extra export and two more nitpicks. Ran it locally and looks good!

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: done

fhlavac (approved):
Looks good! Thank you :tada:

PR #1030, opened 2023-02-15
Define Provisioning manual steps for AWS

ezr-ondrej:
https://issues.redhat.com/browse/RHCLOUD-24246

PR #1032, opened 2023-02-16
Use react-query for bit tidier code

ezr-ondrej:
React query seems quite tidier to me, so I guess it makes sense to use it.
Happy to hear if I'm missing something and we can't actually use it :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "react-intl": "^6.0.5",
+        "react-query": "^3.39.3",
         "react-redux": "^8.0.2",
         "react-router-dom": "^5.3.0",
         "redux": "^4.2.0",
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
I'll fix tests if we decide to use this :)

Hyperkid123:
@ezr-ondrej we need to have a bit more discussions about it. Sources APi uses graphql and there are also axios-based REST clients.

I don't have anything against react-query as such but I don't want to have another data-fetching library in the project. If we want to migrate towards it then we should do the whole app and remove the other libs.

ezr-ondrej:
> If we want to migrate towards it then we should do the whole app and remove the other libs.

Agreed, I guess it stems from ease of use of react-query, I chose to add it, but I guess I could just learn how to use graphQL for my needs in https://github.com/RedHatInsights/sources-ui/pull/1030#discussion_r1108322935

Happy to have a discussion and help with the move to react-query if we decide to do so, or just close this one out :)

Hyperkid123:
As I said, I have nothing against react query I do quite like it, but it's not going to be a trivial migration. We need some planning discussion if it is even worth it atm.

ezr-ondrej:
Happy to help doing so if we ever decide to do it tho :)

PR #1033, opened 2023-02-17
Use application labels consistently

ezr-ondrej:
Always use labelMapper for user facing app names.

### DIFF
--- a/src/components/SourceEditForm/parser/application.js
+++ b/src/components/SourceEditForm/parser/application.js
@@ -5,11 +5,7 @@ import { authenticationFields } from './authentication';
 import { endpointFields } from './endpoint';
 import EditAlert from './EditAlert';
 import ResourcesEmptyState from '../../SourceDetail/ResourcesEmptyState';
-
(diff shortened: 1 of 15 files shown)


src/components/SourceEditForm/parser/application.js
```diff
@@ -56,7 +52,7 @@ export const applicationsFields = (applications, sourceType, appTypes) => [
               id: 'resourceTable.emptyStateDescription',
               defaultMessage: '{applicationName} resources will be added here when created.',
```
ragevou, line 53:
```suggestion
              defaultMessage: 'You will see the {applicationName} resources here after creation.',
```

ezr-ondrej:
Might be completely wrong approach, we might also change the `display_name` on backend and leave the label just for "additional sugger for the label in the wizard"
In which case, happy to close this one out and change it in backend :)

ezr-ondrej:
Probably bit easier approach: RedHatInsights/sources-api-go/pull/638 xD

PR #1042, opened 2023-03-21
RHCLOUD-22885: Update text for Azure EA and MCA scopes

chambridge:
* Update the role association if the scopes are for EA or MCA objects based on testing

### DIFF
--- a/src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
+++ b/src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
@@ -24,6 +24,7 @@ import FormSpy from '@data-driven-forms/react-form-renderer/form-spy';
 const CREATE_AZURE_STORAGE = `${HCCM_DOCS_PREFIX}/html/adding_a_microsoft_azure_source_to_cost_management/assembly-adding-azure-sources#creating-an-azure-storage-account_adding-an-azure-source`;
 const AZURE_CREDS_URL = `${HCCM_DOCS_PREFIX}/html/adding_a_microsoft_azure_source_to_cost_management/assembly-adding-azure-sources#configuring-azure-roles_adding-an-azure-source`;
 const RECURRING_TASK_URL = `${HCCM_DOCS_PREFIX}/html/adding_a_microsoft_azure_source_to_cost_management/assembly-adding-azure-sources#configuring-an-azure-daily-export-schedule_adding-an-azure-source`;
+const AZURE_ROLES_URL = `${HCCM_DOCS_PREFIX}/html/adding_a_microsoft_azure_source_to_cost_management/assembly-adding-azure-sources#configuring-azure-roles_adding-an-azure-source`;
 
 export const ConfigureResourceGroupAndStorageAccount = () => {
   const intl = useIntl();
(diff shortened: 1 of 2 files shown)


fhlavac (dismissed):

fhlavac:
@chambridge looks good, can you please enhance the tests to cover the new conditions added? Thank you

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -124,19 +125,78 @@ const InternalReaderRoleDescription = () => {
...
+      }
+
+      return (
+        <TextContent>
+          <Text component={TextVariants.p}>
+            {intl.formatMessage(
+              {
+                id: 'cost.azure.setupMCAReaderRole',
+                defaultMessage:
+                  'Launch the Azure Portal and give the service principal created above ' + mcaRole + ' role.  {link}',
```
karelhala, line 168: Could be better to use string literal here
```suggestion
                  `Launch the Azure Portal and give the service principal created above ${mcaRole} role.  {link}`,
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -124,19 +125,78 @@ const InternalReaderRoleDescription = () => {
...
+          </Text>
+        </TextContent>
+      );
+    } else {
+      let mcaRole = 'Billing account reader';
+      if (scope.includes('invoiceSections')) {
+        mcaRole = 'Invoice section reader';
+      } else if (scope.includes('billingProfiles')) {
+        mcaRole = 'Billing profile reader';
+      }
```
karelhala, line 159: This can be simplified as
```suggestion
      const mcaRole = scope.includes('invoiceSections') ? 'Invoice section reader' : `Billing ${scope.includes('billingProfiles') ? 'profile' : 'account'} reader`;
```

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
```diff
@@ -124,19 +125,78 @@ const InternalReaderRoleDescription = () => {
...
-          defaultMessage: 'Run the following command in Cloud Shell to create a Cost Management Reader role:',
-        })}
-      </Text>
-      <ClipboardCopy>
-        {`az role assignment create --assignee "${authentication?.username}" --role "Cost Management Reader" --scope "${scope}"`}
-      </ClipboardCopy>
-    </TextContent>
-  );
+  if (scope.includes('billingAccounts')) {
+    if (scope.includes('enrollmentAccounts')) {
```
karelhala, line 129: Just a question about this if, does this change just the titles and texts? If so, we might want to change the texts directly instead of copying same code.

karelhala (commented):
Just a couple nitpicks, but overall looks good!

src/components/addSourceWizard/hardcodedComponents/azure/costManagement.js
chambridge: Sorry, I don't quite understand the comment here.  The if condition changes the text and I wanted to make sure each string was translatable. Open to further suggestions though. I'm definitely not a true FE dev.

karelhala (approved):

PR #1047, opened 2023-03-23
Simpler scope picker

karelhala:


fhlavac (approved):
Thank you @karelhala!

PR #1060, opened 2023-03-31
HCS Cost Management AWS Source Flow

fhlavac:
[RHCLOUD-24156](https://issues.redhat.com/browse/RHCLOUD-24156)

Video with the new flow:

[screen-capture (6).webm](https://user-images.githubusercontent.com/50696716/229121471-c67bc432-d2a1-4416-9145-30280956d136.webm)


src/components/FormComponents/SourceWizardSummary.js
```diff
@@ -27,22 +28,44 @@ import {
...
-            'Make sure to manage permissions for this source in custom roles that contain permissions for Cost Management.',
-        })}
-      </Alert>
+      <Fragment>
+        <Alert
+          variant="info"
+          isInline
+          title={intl.formatMessage({
+            id: 'cost.warningTitleCUR',
+            defaultMessage: "Don't forget additional configuration steps!",
```
karelhala, line 44: Let's use single quotes
```suggestion
            defaultMessage: 'Don't forget additional configuration steps!',
```

src/components/FormComponents/SourceWizardSummary.js
```diff
@@ -89,6 +112,11 @@ export const createItem = (formField, values, stepKeys) => {
     return undefined;
   }
 
+  // do not show fields hidden just for review step
+  if (formField.hideInReview) {
+    return undefined;
```
karelhala, line 117: `undefined` can't be rendered. Wouldn't it be better to return `null`?
```suggestion
    return null;
```

karelhala (commented):
Looking good! Just a few minor nitpicks

src/components/FormComponents/SourceWizardSummary.js
fhlavac: Eslint enforces replacing `'Don\'t forget additional configuration steps!'` with the double quotes.

src/components/FormComponents/SourceWizardSummary.js
fhlavac: we are still applying `filter(Boolean)` at the function's output. Changed the whole function to return `null` everywhere to make it safer

fhlavac: @karelhala it does not probably make sense. The output of the function is not meant to be rendered. 

`filter(Boolean)` is applied to the output and then the results are mapped again into components there:
```
const valuesList = valuesInfo.map(({ label, value }) => (
    <DesctiptionListItem
      key={`${label}--${value}`}
      term={label}
      description={value.toString().length > 150 ? <ValuePopover label={label} value={value} /> : value}
    />
  ));
```

karelhala (approved):

PR #1070, opened 2023-04-06
Adding UID to FinalSteps for adding sources

arivepr:
PR addresses [this JIRA card](https://issues.redhat.com/browse/RHCLOUD-24906)
Adds UUID through a clipboard on the final steps of the 'Add Sources' wizard/modal.
0AA547D2-F601-4872-A578-29A91DBC26F6

FB2889E3-487B-48C5-83C2-0A2DE75E3D2F

### DIFF
--- a/src/components/addSourceWizard/index.js
+++ b/src/components/addSourceWizard/index.js
@@ -144,6 +144,7 @@ const AddSourceWizard = ({
       afterSuccess={afterSuccess}
       sourceTypes={state.sourceTypes}
       activeCategory={activeCategory}
+      sourceUuid={createdSource?.uid}
     />
   );
 };
--- a/src/components/steps/FinishedStep.js
+++ b/src/components/steps/FinishedStep.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import { ClipboardCopy } from '@patternfly/react-core';
 
 import {
   Bullseye,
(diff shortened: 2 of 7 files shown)


src/components/addSourceWizard/index.js
```diff
@@ -144,6 +144,7 @@ const AddSourceWizard = ({
       afterSuccess={afterSuccess}
       sourceTypes={state.sourceTypes}
       activeCategory={activeCategory}
+      sourceUuid={createdSource?.uid}
```
Hyperkid123, line 147: Maybe a typo? `uid -> uuid`? just checking.

src/components/steps/FinishedStep.js
```diff
@@ -24,13 +25,17 @@ const FinishedStep = ({
...
+  uuid,
 }) => (
   <Bullseye>
     <EmptyState variant={EmptyStateVariant.full} className="pf-u-mt-4xl">
       <EmptyStateIcon icon={CheckCircleIcon} color="var(--pf-global--success-color--100)" className="pf-u-mb-0" />
       <Title headingLevel="h2" size="xl" className="pf-u-mt-xl">
         {title}
       </Title>
+      <ClipboardCopy isReadOnly hoverTip="Source UUID" clickTip="Copied" className="pf-u-mt-md">
+        {uuid}
```
Hyperkid123, line 37: Because the prop is not required can we do a check if the UUID exists first before we show the component?

src/components/addSourceWizard/index.js
arivepr: So the object that holds the data indeed has the field as`uid,` I only realized it after i had already named all the variables passed around 😅 

Can have them all be uid if you'd prefer keeping consistency

Hyperkid123: I see. We have a release coming we can fix it later.

Hyperkid123 (approved):

karelhala (approved):
Looking good!

PR #1076, opened 2023-04-26
Anchor Wizard to correct element

shaded-enmity:
By default all modal wizards are anchored to document.body which interferes with Quickstarts by overlaying them. This patch anchors the modal to the main element outside of the Quickstart so that both can be visible at the same time.

### DIFF
--- a/src/components/addSourceWizard/FinalWizard.js
+++ b/src/components/addSourceWizard/FinalWizard.js
@@ -195,7 +195,7 @@ const FinalWizard = ({
     );
   }
 
-  const appendTo = React.useMemo(() => document.querySelector('.pf-c-page.chr-c-page'));
+  const appendTo = React.useMemo(() => document.querySelector('.pf-c-page.chr-c-page'), []);
 
   return (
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

src/components/addSourceWizard/FinalWizard.js
```diff
@@ -195,21 +195,24 @@ const FinalWizard = ({
     );
   }
 
+  const appendTo = React.useMemo(() => document.querySelector('.pf-c-page.chr-c-page'));
```
karelhala, line 198: You are missing dependency array
```suggestion
  const appendTo = React.useMemo(() => document.querySelector('.pf-c-page.chr-c-page'), []);
```

karelhala (commented):
Looking hella good!

src/components/addSourceWizard/FinalWizard.js
shaded-enmity: Good catch, thanks!

karelhala (approved):

PR #1089, opened 2023-05-05
Fix final step width to 58%

karelhala:
### Before
Screenshot from 2023-05-05 15-59-57

### After

Screenshot from 2023-05-05 15-58-35


fhlavac (approved):

fhlavac:
@karelhala unused import

PR #1097, opened 2023-05-31
Remove usage of window.insights

Hyperkid123:


src/components/steps/AmazonFinishedStep.js
```diff
@@ -33,16 +34,16 @@ AwsLink.propTypes = {
...
-
 const AmazonFinishedStep = ({ onClose }) => {
   const intl = useIntl();
+  const { isBeta } = useChrome();
+  const PREFIX = isBeta() ? 'preview/' : '';
+
+  const GOLDIMAGES_HREF = 'https://access.redhat.com/management/cloud';
+  const SUBWATCH_HREF = `/${PREFIX}subscriptions`;
+  const INSIGHTS_HREF = `/${PREFIX}insights`;
+  const COST_HREF = `/${PREFIX}cost-management`;
```
karelhala, line 45: For these we could create [tagget template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates)

```JS
const prefixed = (strings, isBeta) => {
  return `/${isBeta() ? 'preview/' : '';}${strings[1]}`
}
```

We can then use it as
```JS
prefixed`${isBeta}${COST_HREF}`
// or
prefixed`${isBeta}cost-management`
```

PR #1162, opened 2023-09-15
Added RHEL metering to AWS Cost Management Source

radekkaluzik:
https://issues.redhat.com/browse/RHCLOUD-26237

### DIFF
--- a/src/components/FormComponents/PFDivider.js
+++ b/src/components/FormComponents/PFDivider.js
@@ -1,0 +1,16 @@
+import React from 'react';
+
+import { Divider } from '@patternfly/react-core';
+import { useFlag } from '@unleash/proxy-client-react';
+
+const PFDivider = () => {
+  const rhelAws = useFlag('platform.sources.metered-rhel');
+
+  if (!rhelAws) {
+    return null;
+  }
+
+  return <Divider />;
+};
+
+export default PFDivider;
--- a/src/components/FormComponents/RhelAwsCheckbox.js
+++ b/src/components/FormComponents/RhelAwsCheckbox.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+
+import { Checkbox } from '@patternfly/react-core';
+import { useFlag } from '@unleash/proxy-client-react';
+
+const RhelAwsCheckbox = ({ ...props }) => {
+  const { label, input } = useFieldApi(props);
+  const rhelAws = useFlag('platform.sources.metered-rhel');
+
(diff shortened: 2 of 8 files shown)


src/components/FormComponents/PFDivider.js
```diff
@@ -0,0 +1,12 @@
+import React, { Fragment } from 'react';
+
+import { Divider } from '@patternfly/react-core';
+import { useFlag } from '@unleash/proxy-client-react';
+
+const PFDivider = () => {
+  const rhelAws = useFlag('platform.sources.metered-rhel');
+
+  return <Fragment>{rhelAws && <Divider />}</Fragment>;
```
Hyperkid123, line 9:
```jsx
if(!rhelAws) {
  return null
}

return <Divider />
```

Also, I think we might want to make this component slightly generic. But we don't really have custom conditions. You might want to use the divider component even if the flag is false in some other cases. But I think it would need some DDF changes.

src/components/FormComponents/RhelAwsCheckbox.js
```diff
@@ -0,0 +1,32 @@
...
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+
+import { Checkbox } from '@patternfly/react-core';
+import { useFlag } from '@unleash/proxy-client-react';
+
+const RhelAwsCheckbox = ({ ...props }) => {
+  const { label, input } = useFieldApi(props);
+  const rhelAws = useFlag('platform.sources.metered-rhel');
+
+  return (
```
Hyperkid123, line 12: Same as above. Check the flag and return `null` instead of wrapping the actual node in Fragment.

src/components/FormComponents/RhelAwsCheckbox.js
```diff
@@ -0,0 +1,32 @@
...
+          }}
+        />
+      )}
+    </Fragment>
+  );
+};
+
+RhelAwsCheckbox.propTypes = {
+  label: PropTypes.elementType.isRequired,
+  id: PropTypes.elementType.isRequired,
```
Hyperkid123, line 29: Id should never be element type.

src/components/FormComponents/RhelAwsCheckbox.js
```diff
@@ -0,0 +1,32 @@
...
+            input.onChange(checked ? 'rhel' : undefined);
+          }}
+        />
+      )}
+    </Fragment>
+  );
+};
+
+RhelAwsCheckbox.propTypes = {
+  label: PropTypes.elementType.isRequired,
```
Hyperkid123, line 28: I think this should be node rather than element type.

karelhala (approved):

PR #1176, opened 2023-10-31
Add tabs for Integrations Phase 2

fhlavac:
Part of [RHCLOUD-28950](https://issues.redhat.com/browse/RHCLOUD-28950)

Merge after https://github.com/RedHatInsights/notifications-frontend/pull/485

image


src/components/TabNavigation.js
```diff
@@ -50,14 +52,34 @@ const TabNavigation = () => {
...
       />
-      {enableIntegrations && (
-        <Tab
-          eventKey={INTEGRATIONS}
-          title={
-            <TabTitleText>{intl.formatMessage({ id: 'sources.integrations', defaultMessage: 'Integrations' })}</TabTitleText>
-          }
-        />
-      )}
+      {enableIntegrations &&
```
karelhala, line 55: Can we have a fallback to phase 1? If `enableIntegrations` is correct we'll show whole integrations table, if the `enableBreakdown` is true we'll show the list of tabs.

src/pages/Sources.js
```diff
@@ -185,8 +186,8 @@ const SourcesPage = () => {
   const setSelectedType = (selectedType) => stateDispatch({ type: 'setSelectedType', selectedType });
 
   const mainContent =
-    activeCategory === INTEGRATIONS && enableIntegrations ? (
-      <AsyncComponent appName="notifications" module="./IntegrationsTable" />
+    [INTEGRATIONS, COMMUNICATIONS, REPORTING, WEBHOOKS].includes(activeCategory) && enableIntegrations ? (
+      <AsyncComponent appName="notifications" module="./IntegrationsTable" activeCategory={activeCategory} />
```
karelhala, line 190: Same as above, we probably want to fallback to previous implementation if `platform.sources.breakdown` is disabled.

karelhala (commented):
Looking good! Just two minor things.

src/pages/Sources.js
fhlavac: Tested again and I hope it works this way now

src/components/TabNavigation.js
fhlavac: Thank you, this needed an update, done.

karelhala (approved):

PR #1199, opened 2024-01-12
Add exception for sources created by operator

fhlavac:
Fixes [RHCLOUD-27445](https://issues.redhat.com/browse/RHCLOUD-27445)

image

Problem:
"The CostManagementMetricsOperator can create OCP sources via the API. The operator uses a relatively old API flow, but the source creation and integration with cost-management works just fine. However, the cluster-id for those sources is missing in the Sources UI."

Solution:
Added an exception to the process of generating an edit form which shows the the previously not visible field in a disabled state together with a helper text.

### DIFF
--- a/src/components/SourceEditForm/parser/authentication.js
+++ b/src/components/SourceEditForm/parser/authentication.js
@@ -1,9 +1,11 @@
 import React from 'react';
 import get from 'lodash/get';
 import validatorTypes from '@data-driven-forms/react-form-renderer/validator-types';
+import { COST_MANAGEMENT_APP_NAME } from '../../../utilities/constants';
 import { FormattedMessage } from 'react-intl';
 import { doLoadRegions } from '../../../api/doLoadRegions';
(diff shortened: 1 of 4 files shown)


karelhala (approved):

src/components/SourceEditForm/parser/authentication.js
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 import get from 'lodash/get';
 import validatorTypes from '@data-driven-forms/react-form-renderer/validator-types';
+import { COST_MANAGEMENT_APP_NAME } from '../../../utilities/constants';
 import { FormattedMessage } from 'react-intl';
 import { doLoadRegions } from '../../../api/doLoadRegions';
 import hardcodedSchemas from '../../../components/addSourceWizard/hardcodedSchemas';
+import { MANUAL_CONFIGURATION } from '../../constants';
```
fhlavac, line 8:
```suggestion
```

PR #1216, opened 2024-07-04
Fix license

ezr-ondrej:
### Description

Fixes LICENSE to be in a standard file name, bumps tsc-transform-imports.


ezr-ondrej:
@fhlavac can we get this in to fix the pipeline? :pray:

PR #1227, opened 2024-08-08
Enable default tab to be overview when FF is enabled

karelhala:
### Description

Right now when feature flag to enable overivew is set to true the overview page is not opened by default. This PR fixes such issue.

### DIFF
--- a/src/test/pages/Sources.test.js
+++ b/src/test/pages/Sources.test.js
@@ -125,6 +125,7 @@ describe('SourcesPage', () => {
 
     store = getStore([], {
       user: { writePermissions: true },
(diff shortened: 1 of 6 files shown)


src/test/pages/Sources.test.js
```diff
@@ -890,6 +891,8 @@ describe('SourcesPage', () => {
     it('should call onFilterSelect with applications', async () => {
       const user = userEvent.setup();
 
+      console.log(store, 'this is store!');
+
```
fhlavac, line 895:
```suggestion
```

fhlavac (approved):

PR #1341, opened 2025-08-06
fix(source-tiles): update clouds logos

avitova:
### Description
Hi. We found out in the image builder services, that the asset's URL has changed. I figured out you might like knowing about this, too. :)
https://github.com/osbuild/image-builder-frontend/pull/3505
image


epwinchell:
@avitova Thanks for catching this!

avitova:
Thanks!:)

PR #1489, opened 2026-06-15
feat(RBAC): Add v1 wildcard fallback for v2 orgs

aferd:
For v2 orgs, load BOTH Kessel v2 permissions AND v1 Chrome API permissions, then combine them in Redux using OR logic. This ensures Org Admins and legacy roles with wildcard permissions (integrations:*:*, integrations:endpoints:*) maintain access when their org is migrated to v2.

Why: Kessel v2 does not currently support wildcard permission expansion. Without this fallback, users with wildcard permissions would lose access when their org is migrated to v2.

Implementation:
- PermissionsChecker.tsx: Load v1 permissions even for v2 orgs
- reducer.js: Combine permissions with OR logic in Redux state

Temporary workaround: Should be removed once EITHER:
1. Kessel v2 supports wildcard permission expansion, OR
2. All wildcard permissions are migrated to explicit Kessel workspace relations

Related: insights-chrome PR #3362, notifications-frontend PR (same branch)
Issue: RHCLOUD-48396

### Description

description text...

[RHCLOUDXXXX](https://issues.redhat.com/browse/RHCLOUDXXXX)

---

#### After:

---

### Checklist ☑️
- [ ] PR only fixes one issue or story 
- [ ] Change reviewed for extraneous code 
- [ ] UI best practices adhered to 
- [ ] Commits squashed and meaningfully named 
- [ ] All PR checks pass locally (build, lint, test, E2E)

##
- [ ] _(Optional) QE: Needs QE attention (OUIA changed, perceived impact to tests, no test coverage)_
- [ ] _(Optional) QE: Has been mentioned_
- [ ] _(Optional) UX: Needs UX attention (end user UX modified, missing designs)_
- [ ] _(Optional) UX: Has been mentioned_
##

### DIFF
--- a/src/components/PermissionsChecker.tsx
+++ b/src/components/PermissionsChecker.tsx
@@ -49,7 +49,6 @@ const PermissionsChecker: React.FC<PermissionsCheckerProps> = ({ children }) => 
   // Effect to load v1 integrations permissions (for v1 orgs only)
   useEffect(() => {
     if (!isV2Org) {
-      console.log('[SOURCES RBAC DEBUG] v1 org - loading Chrome API permissions');
       Promise.all([
         dispatch(loadIntegrationsEndpointsPermissions(getUserPermissions)),
         dispatch(loadIntegrationsReadPermissions(getUserPermissions)),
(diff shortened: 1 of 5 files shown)


src/components/PermissionsChecker.tsx
```diff
@@ -46,13 +46,25 @@ const PermissionsChecker: React.FC<PermissionsCheckerProps> = ({ children }) =>
...
+    if (!isV2Org) {
+      Promise.all([
+        dispatch(loadIntegrationsEndpointsPermissions(getUserPermissions)),
+        dispatch(loadIntegrationsReadPermissions(getUserPermissions)),
+      ]);
+    }
+  }, [isV2Org, getUserPermissions, dispatch]);
+
+  // Effect to load integrations permissions for v2 orgs (Kessel + v1 fallback)
   useEffect(() => {
```
riccardo-forina, line 60: Yeah, I don't see why we need 2 useEffect here can't we simplify in a single one? It's kinda hard to follow the flow this way

aferd: Yeah, this was to fix a race condition, v1 orgs re-run when Kessel state changes. I could extract this into a custom hook to simplify the flow a bit.

riccardo-forina: I think it's ok, we'll need to rework this repo heavily anyway

riccardo-forina (approved):

PR #1503, opened 2026-07-21
fix(RHCLOUD-49447): remove per-org v2 gating, use global Kessel access checks

aferd:
## Summary

- Remove `platform.rbac.workspaces` Unleash flag from `PermissionsChecker` — Kessel v2 access checks now run globally instead of per-org
- v1 Chrome API permissions load unconditionally as the always-on baseline; Kessel v2 supplements via Redux OR logic when a workspace is available
- Fix `KesselRbacAccessProvider` bug: empty `resource_id` sent to `/api/kessel/v1beta2/checkself` when workspace fetch fails (HTTP 400)
- Fix `KesselRbacAccessProvider` stuck `isLoading` when workspace is unavailable — SDK hooks stay in `loading: true` initial state forever

## Context

Under dual-write, v2 permission checks return the same results as v1 for all orgs, so per-org gating via `platform.rbac.workspaces` was unnecessary. Per clarification from Keith Walsh and Josejulio Martinez Magana, the workspace flag should only gate UX-level decisions, not API call routing.

Ref: [RHCLOUD-49447](https://redhat.atlassian.net/browse/RHCLOUD-49447)

## Changes

| File | Change |
|------|--------|
| `PermissionsChecker.tsx` | Remove `useFlag('platform.rbac.workspaces')`, load v1 unconditionally, dispatch Kessel only when `workspaceId` available |
| `KesselRbacAccessProvider.tsx` | Return `undefined` workspace when no ID (prevents 400), fix `isLoading` stuck state |
| `PermissionsChecker.test.tsx` | Rewrite tests for global approach (removed Unleash mock) |
| `kesselActions.ts`, `permissionMapper.ts` | Doc comment updates (removed "v2 orgs" language) |
| `README.md` | Updated architecture docs to reflect global approach |

## Test plan

- [x] All 147 test suites pass (1063 tests)
- [ ] Manual: org admin can create integrations (all tabs)
- [ ] Manual: non-admin sees correct tabs based on permissions
- [ ] Manual: no Kessel 400 errors in network tab when workspace unavailable
- [ ] Manual: FedRAMP/no-Kessel environment falls back to v1 gracefully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[RHCLOUD-49447]: https://redhat.atlassian.net/browse/RHCLOUD-49447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/PermissionsChecker.tsx
+++ b/src/components/PermissionsChecker.tsx
@@ -1,7 +1,6 @@
 import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
-import { useFlag } from '@unleash/proxy-client-react';
 import { ThunkDispatch } from 'redux-thunk';
 import { AnyAction } from 'redux';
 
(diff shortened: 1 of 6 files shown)


riccardo-forina:
Closing — RHCLOUD-49447 has been closed as Won't Do. Notification apps are in the FedRAMP boundary, so we cannot drop v1 support. Removing per-org gating would break the service in-boundary once promoted to GovCloud. See RHCLOUD-49549 for the replacement approach (v1/v2 coexistence at FEO config level).

